PR #1472, opened 2014-08-01
[Activemq] Invoking `String.unescapeHTML()` in administrative commands fixes #1388

### LINKED ISSUE #1388: suggest popup in ActiveMQ
Suggest popup used on moving messages to another queue  escapes queue names with uri syntax for example "sales/sales-pending-orders" =>  "sales&#x2f;sales-pending-orders" and this is not only in view because if you try to move message to such a queue it creates another one with malformed name

paoloantinori:
fixes #1388


woczkowski:
Unfortunately seems we have regression on this issue. In 1.4.68 input box for destination was replaced by combo and names of destination are not un-escaped again :( now it is even worse because you cannot write destination name by hand as it is combo.

Issue #1202, opened 2014-04-04
Slightly weird rendering 


jhberges:
![capture](https://cloud.githubusercontent.com/assets/1487862/2614836/8d704810-bbef-11e3-9664-5fe90327547a.PNG)

Based on XML source (that is; the code is JavaDSL, but XML from JMX I presume):

```
<split parallelProcessing="true" streaming="true">
    <setHeader headerName="SourcePlatform">
        <simple>${body[SOURCE_PLATFORM]}</simple>
    </setHeader>
    <process/>
    <choice>
        <when>
            <simple>${body[MIGRATION_STATE_NAME]} == 'SCHEDULED'</simple>
            <to uri="direct:classifyCpe"/>
        </when>
        <when>
            <simple>${body[MIGRATION_STATE_NAME]} == 'FAIL_PROVISIONING'</simple>
            <to uri="direct:migrateCpe"/>
        </when>
        <when>
            <simple>${body[MIGRATION_STATE_NAME]} == 'FAIL_CHECK'</simple>
            <to uri="direct:checkOperationalState"/>
        </when>
        <otherwise>
            <log message="... and why are you here, little entry? ${body[CPE_MIGRATION_ID]}" loggingLevel="WARN"/>
        </otherwise>
    </choice>
</split>
<log message="Done migration run"/>
```

Shouldn't there be a joining of the graph after the choice() branches, ending i the "Log" with count 14 on it?
The code appears to _work_ that way, anyhow... ;-)

tadayosi:
It's an issue for Hawtio v1. Let's raise a new issue if we make sure it also happens to Hawtio v2.

PR #1798, opened 2014-12-04
Camel Diagrams: Improved rendering. See #1202

paoloantinori:
Improves rendering of Camel diagrams for nodes with multiple exits. Currently supports only `choice` and `multicast`

See #1202


davsclaus:
Lets close as this is old now

PR #1812, opened 2014-12-08
Hawtio and Keycloak integration

mposolda:


mposolda:
Anyone have some idea about this PR? Just a reminder, I don't want to push hard for it... 
I would need anyway to rebase as it's couple of hundreds commits behind current hawtio master. Also would need to upgrade keycloak version. But still I would appreciate some feedback..

davsclaus:
I for sure think we would love to get keycloak integrated.

I would refer to @jstrachan and @gashcrumb about the best timing. As this PR (I think) touches a bit in the current and a bit crusty auth code I haven't dared to merged it.

gashcrumb:
Took a look at the PR; it touches a lot, but yeah, I'd love for keycloak integration, especially for 2.x, where I've currently avoided moving any of the 1.x auth code over yet.  Can you rebase whenever you get a chance and then I say we merge it :-)

mposolda:
Sure, Thanks for the feedback!. The plan is to release keycloak 1.1.0.Final this week. Once we have it, I will rebase and update PR to use KC 1.1.0.Final at the same time. Will let you know then...

mposolda:
I've rebased with current maste and updated PR to newly released Keycloak 1.1.0.Final . 

I've also changed code a bit to avoid touch existing stuff as much as possible. So PR is especially about adding new stuff and not much changes. The major change to existing code is likely adding "postLogoutTasks", which is needed so Keycloak is able to properly finish logout on it's side. 

However when KC integration is not enabled, postLogoutTasks are empty and behaviour is same like now. Generally, there shouldn't be any change in behaviour when KC integration is disabled.

gashcrumb:
@mposolda many thanks!

PR #1892, opened 2015-02-02
KC integration improvements

mposolda:
- Support for links to Keycloak admin console and account management (if KC integration enabled)
- Fix the wildfly issue when AuthenticationFilter and UserServlet could return different values for 'authenticationEnabled'


gashcrumb:
thanks!

PR #1928, opened 2015-03-19
Check for null servlet context ConfigManager attribute

soleger:
You should check for a null ConfigManager before trying to access it to retrieve the 'authenticationEnabled' setting. In my environment (OSGi HTTP service), registering context parameters is disabled so the parameter will always be missing. This should not cause a NullPointerException when initializing the UserServlet.


davsclaus:
Thanks for the PR

PR #1969, opened 2015-06-03
Spring boot support

stalet:
I was missing support for spring-boot out of the box. The hawtio-sample-springboot project did not work, it only shows a login box since none of the servlets was not deployed. 
So i created a spring-boot HawtioConfiguration class that I added to the project and modified the sample-project so that it will work with spring boot.

The HawtioConfiguration class can be reused, so I moved it to the hawtio-spring-boot project. 
This makes the use of Hawtio from a spring-boot application really easy and convenient.

The HawtioConfiguration is the web.xml rewritten for a servlet 3.0 spring-boot container, and must be updated if there is changes to the web.xml in the hawtio-web project.
It works with both jetty and tomcat.

I would like you to add this to the project so that it can be maintained along with the hawtio project.


davsclaus:
Thanks alot for all this works.

donovanmuller:
When will this be ported to Hawtio 2.x and under which sub-project?

PR #1974, opened 2015-06-19
Removed redirect from / since this will take over the root of the project.

stalet:
Removed the redirect from / to hawtio/index.html.

Spring boot users can add the registry.addViewController("/").setViewName("redirect:/hawtio/index.html"); in their own configuration if they need it. For most users this will only be a hassle if they do not want it.


davsclaus:
Thanks for the PR

stalet:
You're welcome!

Issue #1981, opened 2015-06-30
Wrong ObjectName in online Documentation


predic8:
At page:

http://hawt.io/plugins/health/

the example for an objectName is wrong:

org.apache.activemq:Type=Broker,BrokerName=localhost,service=Health

broker and type should be written in lower case like:

org.apache.activemq:type=Broker,brokerName=localhost,service=Health

That make it easier to copy the ObjectName into a Jolokia URL.

davsclaus:
Thanks a good idea. The doc is here
https://github.com/hawtio/hawtio/blob/master/hawtio-web/src/main/webapp/app/health/doc/help.md

You are welcome to work on a PR if you fancy trying that.

PR #1982, opened 2015-06-30
Update help.md

predic8:
Fixes hawtio/hawtio#1981

 Changed the ObjectName to lowerCamelCase so that it will work if you copy it into a Jolokia URL


davsclaus:
Thanks for this PR

PR #2006, opened 2015-08-19
[ENTESB-3844] host identification color in preferences is greyed out …

kevinearls:
…after selection.  @grgrzybek  or @gashcrumb  can you review this and merge if you think it's ok?  I ask because I removed the line marked "Important", but that's in fact the cause of the problem.

I increased the width of the border to 3px because at 1 or 2 the border tends to get lost with darker colors.

### DIFF
--- a/hawtio-web/src/main/webapp/app/themes/css/default.css
+++ b/hawtio-web/src/main/webapp/app/themes/css/default.css
@@ -653,7 +653,7 @@ div.hawtio-form-tabs ul.nav-tabs li.active a {
 }
 
 .color-picker div table tr td div {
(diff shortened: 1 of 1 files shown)


gashcrumb:
A damn, it's due to the 'selected' class there being used for more than one thing.  Might be good to leave that 'selected' class as-is, as taking that background-color rule out will affect other pages, and instead adjust the hawtio color picker directive to use a different class name for the 'selected' item, i.e. something like 'color-picker-selected'.

kevinearls:
Thanks.  I'll fix that and resubmit.

Issue #1985, opened 2015-07-08
Some images links broken when embedded in Spring Boot


nwoolls:
Using the latest [hawtio-springboot](https://github.com/hawtio/hawtio/tree/master/hawtio-springboot) results in broken image links for some icons such as Camel route children. Fixing this requires the following addition to [HawtioConfiguration.addResourceHandlers()](https://github.com/hawtio/hawtio/blob/master/hawtio-springboot/src/main/java/io/hawt/springboot/HawtioConfiguration.java#L39):

```java
registry.addResourceHandler("/img/**").addResourceLocations("classpath:/static/hawtio/img/");
```

davsclaus:
Thanks for spotting. You are welcome to do a PR with this fix.

nwoolls:
I pinged the [author](https://github.com/stalet) of the [recent changes](https://github.com/hawtio/hawtio/pull/1969) to [hawtio-springboot](https://github.com/hawtio/hawtio/tree/master/hawtio-springboot) on Stack Overflow first to see if it's something he was going to keep maintaining. If not I'll submit a PR.

stalet:
I'm currently on a holiday, so feel free to submit a PR, or i will do the change when i'm back in. August.

davsclaus:
Anyone got the time to work on a PR?

stalet:
Sure, ill start working on one.

PR #2008, opened 2015-08-21
Added resourcehandler for /img/**. This commit fixes hawtio/hawtio#1985

stalet:
Fix for hawtio/hawtio#1985


jimmidyson:
Thanks!

Issue #2025, opened 2015-10-22
spring boot example and /perspective redirect


redron:
Hi,
I am trying to run example `hawtio-sample-spingboot` from latest sources, but get stuck with the following issue:

when I type http://localhost:10000/hawtio/index.html in a browser (FF, Chrome) it starts loading, but after showing menu-bar URL get redirected to http://localhost:10000/hawtio/perspective/defaultPage

1 As the result further loading fails, since it tries to reuse URL from browser to locate static resources:

```
[Core] hawtio started
GET http://localhost:10000/hawtio/perspective/app/core/html/layoutFull.html [HTTP/1.1 404 Not Found 3ms]
GET http://localhost:10000/hawtio/perspective/app/perspective/html/defaultPage.html  [HTTP/1.1 404 Not Found 2ms]
```
But correct ones should be:
```
http://localhost:10000/hawtio/app/core/html/layoutFull.html
http://localhost:10000/hawtio/app/perspective/html/defaultPage.html
```

After several hours of research I thought I found some workaround. Adding to Spring HawtioConfiguration class following setting:
```
registry.addViewController("/hawtio/perspective/**").setViewName("redirect:/hawtio/**");
```
2  But this fix leads to syntax error in app.js code. 
```
Stack trace:  <div class="log-stack-trace">
<p>Error: expected expression, got '<'</p>
<p>.globalEval/<@http://localhost:10000/hawtio/app/app.js?a62f0c433b149c2d:1:26928</p>
<p>.globalEval@http://localhost:10000/hawtio/app/app.js?a62f0c433b149c2d:1:26902</p>
...
<p>r@http://localhost:10000/hawtio/app/app.js?a62f0c433b149c2d:15:2105</p>
<p>sb/</v.onreadystatechange@http://localhost:10000/hawtio/app/app.js?a62f0c433b149c2d:15:2740</p>
```
Please advice what might be cause of this redirect issue.

Thanks in advance,
Andrey

pwielgolaski:
I have similar problem, but I noticed that when it is run with 1.5-SNAPSHOT it works ok, when I switch to 1.4.56 there is the same issue that you described.
It seems that release build could build JS code in different way for SNAPSHOT/release build.

pwielgolaski:
I noticed that when I remove this replacement https://github.com/hawtio/hawtio/blob/master/hawtio-springboot/pom.xml#L92
it works fine for release build.

[...]

PR #2026, opened 2015-10-25
Fix for #2025, handle html5 deactivation for uglified JS

pwielgolaski:
There is issue with replacing JS code after it was uglified, so release build does not do html5Mode switch for Spring boot and it fails to start properly. See issue https://github.com/hawtio/hawtio/issues/2025


davsclaus:
Thanks for the PR

PR #2076, opened 2016-01-22
ENTESB-4331 Remove stack traces from jolokia requests

kevinearls:
Hey @gashcrumb and/or @grgrzybek can you review this?  Thanks.

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/RedirectFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/RedirectFilter.java
@@ -1,8 +1,10 @@
 package io.hawt.web;
 
+import java.io.CharArrayWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PrintWriter;
 import java.util.regex.Pattern;
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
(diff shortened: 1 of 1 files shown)


davsclaus:
Hold on merging this until we get a better solution out of the box from Jolokia.

rhuss:
I wonder if more stuff should be removed in case of an error. Typically I add as much context information to error message as possible in order to help maintenance.

davsclaus:
Yeah I guess its a thin line to walk. 

Though maybe if the extended error is turned off, you only return http status code and status line. 

The actual error details are protected on the server side in the server logs.

Then with this configuration jolokia is more secure and may not expose too much details to malicious clients.

rhuss:
Thinking again, I think even exposing the stack trace is not really a security issue, because you need be authenticated anyway to get that far. I mean, if you can get the stack trace by provoking an error when calling a MBean, you can easily get the full thread dump of the whole VM by executing an operation on `java.lang:type=ThreadDump`, too.

Also, all a stack trace reveals is the executing path through the Jolokia agent to the JMX subsystem, which is more or less always the same (so sending the stack trace is probably not very useful anyway), except when an JMX operation is called which jumps into some other code but this is hopefully only used for management, too.

So, if someone can execute Jookia to get a stacktrace he already can do much more severe things.

Neverhteless, I'll add the option we discussed above to Jolokia 1.3.3. Can't harm either ;-)

And for 2.0 I'll add the default to **not** send a stacktrace. makes more sense to be now.

kevinearls:
@rhuss and @davsclaus Thanks!

Issue #1965, opened 2015-05-11
Support a list of predefined jolokia agents 


MarkBramnik:
Adding this issue from stackoverflow question:

http://stackoverflow.com/questions/30151365/hawtio-predefined-list-of-remote-agents

Claus has asked additional questions there, so I'll explain my point of view (hopefully its a right place of doing so):

Our application is comprised of many deployment units residing in different servers in cluster. 
All the applications are written in Java and should have jolokia agent installed

Automation/dev ops would benefit if hawtio could support a server side configuration.
They could generate some kind of XML/properties file in a predefined format during the deployment process and supply that file to hawtio as an external parameter. An Example (possible XML file for 2 servers):

    <farm-configuration>
      <name>myfarm1</name>
      <participants>
          <participant>
              <name>server1</name>
              <connectionString>[HOW_TO_CONNECT_TO_JOLOKIA]</connectionString>
              <security>....</security>  
        <participant>
              <name>server2</name>
              <connectionString>[HOW_TO_CONNECT_TO_JOLOKIA]</connectionString>
              <security>....</security>  
         </participant>      
        </participants>
    </farm-configuration>

So that ui side would just "see" the representation of this file.

Now regarding the packaging (solely IMHO of course). Although I'm coming from the server side world, I believe that repackaging the WAR is not as versatile as an external configuration, because it would make the deployment process easier (just fetch from nexus and don't repackage during deployment) and in theory it should be less 'fragile' when it comes to upgrades in future where possibly the hawtio app structure may be changed. 

Yet another (more advanced) step toward this direction is to support a dynamic configuration (out of URL or at least providing a watchdog on this configuration file). In this case it will be possible to add/remove agents dynamically, which is the most flexible approach I believe.

sparks1372:
This would be extremely useful.  We want to use HawtIO but without this functionality it just isn't feasible.

gashcrumb:
I for sure think we should consider this for hawtio 2.x, have created an issue in the current 2.x [hawtio-java](https://github.com/hawtio/hawtio-java) project.

haninaguib:
Sort of a hack but https://github.com/hawtio/hawtio/pull/2077 allows you to specify the connection settings via url parameters so you can create links that connect automatically to a given agent.

links follow this pattern:
/hawtio/index.html#/jvm/connect?name=xxx&host=xxx&port=xxx&path=xxx&userName=xxx&password=xxx
(where you replace xxx with the desired values)

[...]

PR #2077, opened 2016-01-25
Addresses ticket #1965

haninaguib:
I am planing on integrating hawtio into a monitoring application we have and wanted the ability to create links that would open up hawtio and automatically connect to specific servers being monitored.

With this change I am able to do so by creating links in this format
/hawtio/index.html#/jvm/connect?name={name}&host={host}&port={port}&path={path}&userName={userName}&password={password}

This is very hacky and I am sure one of you could improve it, I have only started looking at hawtio today and javascript/typescrypt is not something I am too familiar with.

### DIFF
--- a/hawtio-web/src/main/webapp/app/jvm/js/connect.ts
+++ b/hawtio-web/src/main/webapp/app/jvm/js/connect.ts
@@ -159,6 +159,21 @@ module JVM {
       angular.extend($scope.currentConfig, $scope.connectionConfigs[$scope.lastConnection]);
       Core.$apply($scope);
     };
+    
+	var autoconnect = $location.search();
+	if (typeof autoconnect != 'undefined' && typeof autoconnect.name != 'undefined') {	
+		var conOpts = Core.createConnectOptions({
+		        scheme: 'http',
+		        host: autoconnect.host,
+		        path: autoconnect.path,
+		        port: autoconnect.port,
+		        userName: autoconnect.userName,
+		        password: autoconnect.password,
+		        name: autoconnect.name
+		});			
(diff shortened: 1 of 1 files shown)


davsclaus:
Yeah this seems not so hacky after all. Though the one thing is that the password would be submitted in clear-text.

haninaguib:
In my use case I dont use username/password (all agents are on unexposed ports within a VPC)
But maybe username & password should come in via a basic auth header instead of url params.

davsclaus:
Yeah though if you omit username and password then whatever auth is used on that remote jvm ought to kick-in, eg show a login dialog if basic auth etc.

hawtio-web/src/main/webapp/app/jvm/js/connect.ts
```diff
@@ -159,6 +159,21 @@ module JVM {
       angular.extend($scope.currentConfig, $scope.connectionConfigs[$scope.lastConnection]);
       Core.$apply($scope);
     };
+    
+	var autoconnect = $location.search();
+	if (typeof autoconnect != 'undefined' && typeof autoconnect.name != 'undefined') {	
+		var conOpts = Core.createConnectOptions({
+		        scheme: 'http',
```
DonatasKozlovskis: I wonder if it would be possible to change this feature to allow to set the scheme as well, since now auto connect link limits to only http connections. It would be good to have scheme parameter in url as well.

davsclaus: yeah sure you are welcome to do a PR

PR #2083, opened 2016-01-27
Mark classes in hawtio-insight-log that are marshaled over jmx as serializable

kwesterfeld:
- LogResults, LogFilter and LogEvent marked as serializable

We need the classes in hawtio-insight-log to be marked serializable, as we use a multi-tier application architecture where the "web tier" runs hawtio, but back-end services do not install the http feature, nor do they install jolokia.  In this case, we use federation provided by the legacy jdmk to marshal instance data over rmi (jmx bridging).  Because of this, installing hub/spoke log monitoring doesn't work, and we have to cobble together a delegation model using openmbeans.

Alternately, marking this class as @MXBean would work and not require each tier to have the classes installed.   This change simply marks the beans serializable.


davsclaus:
Thanks for the PR

PR #2126, opened 2016-04-25
Addresses ticket #1965,  connection scheme setting update

DonatasKozlovskis:
Updated auto-connect feature by adding ability to choose connection scheme.

### DIFF
--- a/hawtio-web/src/main/webapp/app/jvm/js/connect.ts
+++ b/hawtio-web/src/main/webapp/app/jvm/js/connect.ts
@@ -163,7 +163,7 @@ module JVM {
 	var autoconnect = $location.search();
 	if (typeof autoconnect != 'undefined' && typeof autoconnect.name != 'undefined') {	
 		var conOpts = Core.createConnectOptions({
-		        scheme: 'http',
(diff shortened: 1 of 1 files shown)


hawtio-web/src/main/webapp/app/jvm/js/connect.ts
```diff
@@ -163,7 +163,7 @@ module JVM {
 	var autoconnect = $location.search();
 	if (typeof autoconnect != 'undefined' && typeof autoconnect.name != 'undefined') {	
 		var conOpts = Core.createConnectOptions({
-		        scheme: 'http',
+		        scheme: autoconnect.scheme,
```
davsclaus: I wonder if we need to set http as fallback if the scheme is not configured on autoconnect? Or is it always configured?

davsclaus:
Did you see my comment? Are you able to look into this?

DonatasKozlovskis:
feature updated according your comment- http is set as fallback if the scheme is not configured on autoconnect

davsclaus:
Thanks for coming back to this

PR #2140, opened 2016-06-01
ENTESB-5479 - Added support for externalized jolokia-access.xml config

### LINKED ISSUE #2139: Add support for configurable Jolokia advanced config.
See https://issues.jboss.org/browse/ENTESB-5479

```
    It uses now a system property.
    
    ex.
    
    EXTRA_JAVA_OPTS='-Djolokia.policyLocation=file:///home/fuse/my-access.xml'
```

paoloantinori:
It uses now a system property.

ex.
```
EXTRA_JAVA_OPTS='-Djolokia.policyLocation=file:///home/fuse/my-access.xml' bin/fuse debug
```

Fixes #2139

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/JolokiaConfiguredAgentServlet.java
+++ b/hawtio-system/src/main/java/io/hawt/web/JolokiaConfiguredAgentServlet.java
@@ -1,0 +1,94 @@
+package io.hawt.web;
+import org.jolokia.config.ConfigKey;
+import org.jolokia.http.AgentServlet;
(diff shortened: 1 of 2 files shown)


davsclaus:
Can you format the code properly, and add some javadoc to this class what its intention is. And can you change the INFO log where you say something about "not found" as that will confuse/scare users to think something is maybe wrong.

And please add some docs about this in the .md file
https://github.com/hawtio/hawtio/blob/master/docs/Configuration.md

Issue #2075, opened 2016-01-22
queue names should be clickable


phavekes:
It would really make hawtio easier to use if the queue names on the starting page (activemq -> queue -> attrbutes) are clickable, and behave the same as selecting them in the left tree view (take you to the 'browse' view)

davsclaus:
Yeah as it requires the JMX plugin to show that attributes table, it would need to have some kind of onClick callback hook, so we can add that for various plugins. So AMQ can browse the queue, Camel can show the route diagram or something

PR #2156, opened 2016-06-20
Pull Request for #2075

guillaumeterral:
The idea is that each plugin can register onClickHandler on specific table the same way they are declaring columns and then when a click event happend the hawtio-simple-table will propagate it to the plugin. This is used to fix issue #2075  for ActiveMQ and also for Camel plugin for routes/components/endpoints/context


gashcrumb:
Looks good to me, thanks!

PR #2163, opened 2016-06-26
Sample plugin improved

tsirkin:
I was straggling for some time with creation of my own plugin.

Now what I wanted was:
1. I wanted the plugin to be statically linked since 
    a. Docs says it is the simplest way to do it.
    b. I want the hawtio & the plugin to run at the same jvm, so I can debug them both through my IDE.
2. I want them both to be run & reloaded with simple command (mvn jetty:run) .
3. I want the static files to be reloaded using jetty running in background.

Now, the example plugin runs as separate app. , so it does not fit (I want to be able to debug the main hawtio).
The sample plugin does not help as it is not runnable at all, just a simple placeholder.
Without any actual code.

I managed,
But I think that making the sample to have more functionality would actually be a great help for somebody like me.

I have patched the sample plugin to do all that .
Could somebody review it?
Thanks
Evgeny


davsclaus:
Thanks for the PR

PR #2165, opened 2016-06-30
ENTESB-5219 - Make RBACRestrictor always use canInvoke(String, String, String[])

tadayosi:
A small improvement to #2153.


tadayosi:
See also: https://issues.jboss.org/browse/ENTESB-5775

PR #2183, opened 2016-08-02
ENTESB-5891 - Viewer has admin rights after connecting to another Fuse

tadayosi:
https://issues.jboss.org/browse/ENTESB-5891


davsclaus:
ok to test

PR #2199, opened 2016-09-16
ENTESB-6104 keycloak.js adapter is loaded from keycloak server rather…

mposolda:
… than from bower


mposolda:
Added another commit to improve keycloak and hawtio integration. Note: Changes in non-keycloak related java classes are just to ensure that some functionality can be reused by Keycloak related classes (ie. protected methods, static methods etc)

PR #2203, opened 2016-09-26
Keycloak integration with hawtio on EAP

mposolda:
- Test and update instructions for Keycloak with hawtio on EAP 7
- Upgrade instructions to Keycloak 2.2.1, which has fix for nasty javascript bug
- Better error handling on refreshing token or when keycloak server not available


gashcrumb:
Thanks!

PR #2214, opened 2016-11-02
Updated keycloak type definitions

klinki:
Updated KeyCloak type definitions based on manual: https://keycloak.gitbooks.io/securing-client-applications-guide/content/topics/oidc/javascript-adapter.html


klinki:
This might be related to TypeScript version. Which version do you use?

davsclaus:
Yeah hawtio v1 uses an older TS version. @gashcrumb do you remember?

gashcrumb:
Yeah, it's an ancient version, this is on grunt-typescript 0.4.0 which is typescript 1.1.0-1, so definitely need to stick with typescript 1.x compatible definitions.

davsclaus:
[merge]

davsclaus:
Thanks it compiles now

PR #2241, opened 2016-12-13
[ENTESB-6264] OSGi - Declarative Services - Activate/Deactivate butto…

oscerd:
…n on service individual page looks like do nothing


oscerd:
ok to test

PR #2260, opened 2017-01-20
Problem with simpleDataTable.isSelected

skarsaune:
I'm working on a plugin that uses simpleDataTable. It gives javascript errors as row.entity is attempted passed as a function to array.some(...). 

By using similar logic as in reSelectedItems the problem appears to be fixed. Have run both Camel and Quartz plugins without experiencing issues. 

The use of 's.index' appears a bit suspect, but works in my case as the primaryKeyFn only relies on the entity.


tadayosi:
Looks good to me. Thanks for your contribution!

PR #2270, opened 2017-02-27
Fix #2138 - Hawt.io embedded in spring boot app cannot use different web context

### LINKED ISSUE #2138: Hawt.io embedded in spring boot app cannot use different web context
I have spring boot application which is packaged as WAR file. This application contains embedded hawt.io version 1.4.65 . Hawt.io has beed embedded following https://github.com/hawtio/hawtio/tree/master/hawtio-sample-springboot .

Everything is working correctly when i deploy application to Tomcat ROOT (/) . But when i deploy application to specific web context(for example /myapp) i am not able to access hawt.io GUI anymore. It every time forwards me to login (please note that i have disabled authentication and no login is required when application is deployed in /).

vitaliylevashin:
RedirectFilter
--fix for custom context root
AuthenticationFilter
--fix for a couple of system properties

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/AuthenticationFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/AuthenticationFilter.java
@@ -79,10 +79,10 @@ public class AuthenticationFilter implements Filter {
 
         // JVM system properties can override always
         if (System.getProperty(HAWTIO_AUTHENTICATION_ENABLED) != null) {
-            configuration.setEnabled(Boolean.getBoolean(System.getProperty(HAWTIO_AUTHENTICATION_ENABLED)));
+            configuration.setEnabled(Boolean.getBoolean(HAWTIO_AUTHENTICATION_ENABLED));
         }
         if (System.getProperty(HAWTIO_NO_CREDENTIALS_401) != null) {
-            configuration.setNoCredentials401(Boolean.getBoolean(System.getProperty(HAWTIO_NO_CREDENTIALS_401)));
+            configuration.setNoCredentials401(Boolean.getBoolean(HAWTIO_NO_CREDENTIALS_401));
         }
         if (System.getProperty(HAWTIO_REALM) != null) {
             configuration.setRealm(System.getProperty(HAWTIO_REALM));
--- a/hawtio-system/src/main/java/io/hawt/web/AuthenticationFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/AuthenticationFilter.java
@@ -79,10 +79,10 @@ public class AuthenticationFilter implements Filter {
 
         // JVM system properties can override always
         if (System.getProperty(HAWTIO_AUTHENTICATION_ENABLED) != null) {
-            configuration.setEnabled(Boolean.getBoolean(HAWTIO_AUTHENTICATION_ENABLED));
+            configuration.setEnabled(Boolean.getBoolean(System.getProperty(HAWTIO_AUTHENTICATION_ENABLED)));
         }
         if (System.getProperty(HAWTIO_NO_CREDENTIALS_401) != null) {
-            configuration.setNoCredentials401(Boolean.getBoolean(HAWTIO_NO_CREDENTIALS_401));
+            configuration.setNoCredentials401(Boolean.getBoolean(System.getProperty(HAWTIO_NO_CREDENTIALS_401)));
         }
         if (System.getProperty(HAWTIO_REALM) != null) {
             configuration.setRealm(System.getProperty(HAWTIO_REALM));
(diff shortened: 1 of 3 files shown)


hawtio-system/src/main/java/io/hawt/web/AuthenticationFilter.java
```diff
@@ -79,10 +79,10 @@ public void init(FilterConfig filterConfig) throws ServletException {
 
         // JVM system properties can override always
         if (System.getProperty(HAWTIO_AUTHENTICATION_ENABLED) != null) {
-            configuration.setEnabled(Boolean.getBoolean(HAWTIO_AUTHENTICATION_ENABLED));
+            configuration.setEnabled(Boolean.getBoolean(System.getProperty(HAWTIO_AUTHENTICATION_ENABLED)));
```
tadayosi: This is not really a bug. Instead, the proposed one has a bug. See the javadoc for `Boolean#getBoolean()`.

hawtio-system/src/main/java/io/hawt/web/AuthenticationFilter.java
```diff
@@ -79,10 +79,10 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
         }
         if (System.getProperty(HAWTIO_NO_CREDENTIALS_401) != null) {
-            configuration.setNoCredentials401(Boolean.getBoolean(HAWTIO_NO_CREDENTIALS_401));
+            configuration.setNoCredentials401(Boolean.getBoolean(System.getProperty(HAWTIO_NO_CREDENTIALS_401)));
```
tadayosi: This is not really a bug. Instead, the proposed one has a bug. See the javadoc for `Boolean#getBoolean()`.

hawtio-system/src/main/java/io/hawt/web/RedirectFilter.java
```diff
@@ -42,7 +42,7 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
     }
 
     private void process(HttpServletRequest request, HttpServletResponse response, FilterChain chain) throws IOException, ServletException {
-        String uri = request.getRequestURI();
+        String uri = request.getPathInfo();
```
tadayosi: What kind of bug does it fix?  Can you first file an issue and explain what bug you've found before submitting a pull req?

vitaliylevashin:
Hi Tadayosi,

It seems that I was in a rush. I agree on AuthenticationFilter - it was mistake of mine. 
Regarding context root problem - I've found related open issue - https://github.com/hawtio/hawtio/issues/2138.
I've applied suggested fix, which allowed me to open hawtio console while using custom context root. Unfortunately console still fails to connect to local jolokia agent. It seems that link to jolokia is generated without any regard to context root. 
For example:
my context root is "test"
app is accessible at localhost:8087/test
jolokia endpoint as available at http://localhost:8087/test/jolokia
with applied fixes I can now access console localhost:8087/test/hawtio/index.html
hawtio fails to show me current state of local jvm and in the network view of browser's console I can see 404 code for requests:
1) http://public:biscuit@localhost:8087/test/hawtio/jolokia
2) http://public:biscuit@localhost:8087/jolokia
Maybe there is some workarounds I can use in order to specify proper url (like some system property or, any other kind of configuration)?

tadayosi:
Thank you for the contribution. However, closely looking at #2138 I don't think the fix suggested at it is not a correct one. I suspect that's why you still see issues even after applying the patch.

hawtio-system/src/main/java/io/hawt/web/SessionExpiryFilter.java
```diff
@@ -84,7 +84,7 @@ private void process(HttpServletRequest request, HttpServletResponse response, F
         }
         HttpSession session = request.getSession(false);
         boolean enabled = (boolean) context.getAttribute("authenticationEnabled");
-        String uri = request.getRequestURI();
+        String uri = request.getRequestURI().substring(request.getContextPath().length());
```
tadayosi: This is not correct.

hawtio-system/src/main/java/io/hawt/web/RedirectFilter.java
```diff
@@ -42,7 +42,7 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
     }
 
     private void process(HttpServletRequest request, HttpServletResponse response, FilterChain chain) throws IOException, ServletException {
-        String uri = request.getRequestURI();
+        String uri = request.getRequestURI().substring(request.getContextPath().length());
```
tadayosi: This is not correct, either. See how `myContext` is derived at line 58.

tadayosi:
@vitaliylevashin Let's discuss your issue at #2138. A pull req is not a good place to troubleshoot further, so I'm closing it.

PR #2292, opened 2017-04-19
fix of escaping problem as regeression of #1388

woczkowski:


tadayosi:
Thanks for the contribution!

Issue #2289, opened 2017-04-14
Terminal plugin does not work on Karaf 4.1.x


davsclaus:
They changed the api again, and you can find some error in the stacktrace

![screen shot 2017-04-14 at 4 07 25 pm](https://cloud.githubusercontent.com/assets/477100/25045417/91a9ba6a-212c-11e7-9d40-6858c56c1500.png)

davsclaus:
```
2017-04-14 16:08:30,269 | INFO  | qtp761921009-146 | AuthenticationFilter             | 58 - io.hawt.hawtio-karaf-terminal - 1.5.0 | Failed to invoke action null due to:
java.security.PrivilegedActionException
	at java.security.AccessController.doPrivileged(Native Method) ~[?:?]
	at javax.security.auth.Subject.doAs(Subject.java:422) [?:?]
	at io.hawt.web.AuthenticationFilter.executeAs(AuthenticationFilter.java:224) [58:io.hawt.hawtio-karaf-terminal:1.5.0]
	at io.hawt.web.AuthenticationFilter.doFilter(AuthenticationFilter.java:162) [58:io.hawt.hawtio-karaf-terminal:1.5.0]
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1691) [92:org.eclipse.jetty.servlet:9.3.14.v20161028]
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:582) [92:org.eclipse.jetty.servlet:9.3.14.v20161028]
	at org.ops4j.pax.web.service.jetty.internal.HttpServiceServletHandler.doHandle(HttpServiceServletHandler.java:72) [111:org.ops4j.pax.web.pax-web-jetty:6.0.3]
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143) [91:org.eclipse.jetty.server:9.3.14.v20161028]
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:548) [89:org.eclipse.jetty.security:9.3.14.v20161028]
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:226) [91:org.eclipse.jetty.server:9.3.14.v20161028]
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1180) [91:org.eclipse.jetty.server:9.3.14.v20161028]
[...]

PR #2295, opened 2017-04-22
#2289  Terminal plugin does not work on Karaf 4.1.x

yogeshrao:
Upgraded hawtio-karaf-terminal plugin to Jline version to 3.2.0 available as part of karaf 4.1.1 version


tadayosi:
Thanks for the pull req.

Issue #2294, opened 2017-04-21
Regexp support for proxyWhitelist


fabienrousseau:
Being able to whitelist hosts based on a regexp can prove very useful when there is a high number of servers to connect too (and moreover when those servers are in the cloud, and can frequently change).

fabienrousseau:
I've prepared a branch which allows to pass regexp (by prefixing it by "r:" to avoid transforming hostnames to regexp) and added a unit test.
Note: the method ProxyDetails.isAllowed should probably be in the ProxyWhitelist. An implementation of ProxyWhitelist based on regexp could cache/reuse the Pattern objects, nevertheless, it's not on a critical path so this is probably not a problem.

I'll push it on my fork in a day or 2.

fabienrousseau:
If the pull request needs some rework, I'm keen on doing/helping it (with some guidance on how to do it).

PR #2298, opened 2017-04-24
#2294 Add whitelist regex support

fabienrousseau:


tadayosi:
The whitelist is frequently checked when the proxy is used, so I agree that the `Pattern` object should be cached. I think `ProxyWhitelist` should have the cache. What about adding `protected List<Pattern> regexWhitelist` to `ProxyWhitelist` and adding another method `isAllowed(List<Pattern> regexWhitelist)` to `ProxyDetails`?  `regexWhitelist` should be read-only so I guess there's no need to use a thread-safe collection.

tadayosi:
Please also don't forget to update https://github.com/hawtio/hawtio/blob/master/CHANGES.md. This enhancement is worth noting there.

fabienrousseau:
I agree that `ProxyWhiteList` should have the cache. I will also add the `isAllowed` method to `ProxyDetails` (I tend to avoid doing so in general because it adds code in `ProxyDetails` class, and this code should probably reside in `ProxyWhiteList` only).

Good catch for the CHANGES.md file.

I think I can do the changes by the end of this week.

tadayosi:
Thanks!

fabienrousseau:
Thanks for this awesome tool and your reactivity.

Issue #2299, opened 2017-04-24
Add option to listen on specific interface when using embedded hawtio


fabienrousseau:
Currently, it's not possible to restrict hawtio to listen on a specific network interface.

We would like hawtio to listen on localhost only (because it will be setup behind a proxy).

I will submit a simple PR soon which will:
 - add an "host" option (similar to "port" option which already exists)
 - by default, it listen on "0.0.0.0" (old behaviour)

PR #2300, opened 2017-04-24
#2299 add host option for embedded hawtio

fabienrousseau:


tadayosi:
Thanks for the contribution!

PR #2302, opened 2017-04-28
Plugin Idea: java.lang.Runtime plugin

skarsaune:
I have been thinking about a simple plugin to summarise and visualise basic information about the java process, that is a bit simpler for users less familiar with java.lang.Runtime and java System Properties. This may be useful when you have a lot of tabs to different virtual machines:

Main points:
-  Expandable sections with intuitive headings
- Filterable and sortable system properties
- Clicking on values copies to clipboard

Input is very welcome:
 
1. Does this justify a tab of its own? Is there somewhere else it could fit in?
1. Is there other relevant contents that should be included in a plugin like this?
1. Any input on headings , titles , order etc?

Illustration:

image


davsclaus:
Looks great.

I wonder if we should maybe in the future have a panel to show some summary of current values of
- memory usage
- cpu usage
- io usage (open file handles)
- thread usage (we do have threads plugin, but maybe a quick summary is nice here too)

You can find insipiration in the good old jconsole tool in its VM Summary tab and a few of its other tabs.

skarsaune:
I guess this need can be met by selecting the various numeric attributes as graphs and placed in the dashboard view, but it requires some effort by the user to customise the view. 
I could try to expand by adding a "Key metrics" section inspired by the VM Summary in tabular form. I suppose the most valuable are listed here:    

image

davsclaus:
Yeah those or some of those stats from JConsole would be great.

skarsaune:
Sure. I will try to include some more relevant attributes.

skarsaune:
All right, take 2: 

Have been looking for quick-wins based on tabs in jconsole. This can be realized by adding a key metrics section with a few selected attributes:

image
The links will bring you to more detailed info elsewhere in hawt.io , MBeans or dedicated views. 

As this involved picking up more values from the OperatingSystem, I also put in some relevant details for overview:

image

Next steps: Should probably reorganize into 2 controllers as it deals with quite a few beans at the moment.

What about naming? Does Runtime still cover it ? JVM is already used by the connect tab.

davsclaus:
Yeah I think Runtime is a good name. 

And it sounds like a good idea with maybe have some sub tabs to organize the information into different relevant tabs.

skarsaune:
Thanks for input. Will work on splitting it up.

skarsaune:
Now split into 3 distinct subtabs:
image
image
image

davsclaus:
Yay that looks great

davsclaus:
@skarsaune let us know when you think this PR is so much ready that we can merge this

skarsaune:
Complete as far as I can see. Have no improvements in mind.

davsclaus:
@tadayosi lgtm, wonder if you can take a look, and if so get it merged.

And then we can look at doing a new 1.5.2 release to get this out to the users

oscerd:
Since we are preparing a new release, maybe we can merge this one. It looks good to me.

davsclaus:
@oscerd yeah go ahead

oscerd:
Thanks @skarsaune

Issue #2318, opened 2017-05-18
Class histogram viewer / heap: 404 when running through proxy, bug in parsing of classnames


skarsaune:
2 issues discovered in class histogram viewer:
- may give 404 when running through proxy. This is caused by the jolokia exec request being passed as an object, which in represented as a GET request with the request encoded into the url that is somehow not correct. This is easily by wrapping the request object in an array which is handled as a POST request with the request then being sent as payload.
- Too eager parsing of array types. Causes a problem when encountering classes in the default package starting with a capital L, for instance LessAssetPipelineGrailsPlugin. Solved by more criteria in parsing method.

PR: #2320

PR #2320, opened 2017-05-18
Diagnostics (Class histogram) : #2318 : use POST to get class histogram. Fix parsing of Array member

skarsaune:
types. 
Remove redundant code.
+Ensure that edited flags are updated in GUI.


tadayosi:
Thanks!

Issue #2316, opened 2017-05-15
Upgrade IDE plugin to use REST api to open files in IDEA


skarsaune:
Problems:

1. The IDE plugin tries to resolve filenames to absolute files by the basedir and a TODO pointing to local storage (ideHelpers.ts:34). This does not seem to work as intended. 
2. The IDE plugin attempts to call XmlRpc to IDEA, this simply gives java.net.SocketException: Unexpected end of file from server

Investigation:
IDEA also exposes a REST interface to open files, that seems much simpler, and I have played around with in Postman:
https://github.com/JetBrains/intellij-community/blob/master/platform/built-in-server/src/org/jetbrains/ide/OpenFileHttpService.kt

for instance GET http://localhost:63342/api/file?file=src/main/java/io/hawt/HawtioContextListener.java
Successfully opens the file in IDEA 2017.1. 
The same goes for absolute file path.
Unqualified file name and file name only qualified by Java package does not open the file.
Return value is OK 200 in all cases.
Alternatively a POST json payload can be tested out, whether it gives better feedback. 
In addition one could check with the folks at JetBrains whether the service could be enhanced to figure out the exact file with less input.

Proposed solution:
1. Remove or fix unfinished code that attempts to resolve to absolute file name.
2. Change to use REST api for opening files

UPDATE: One can probe the REST interface by calling GET /api/file, the downside is that the user may get an extra warning confirm inside IDEA about another process attempting to open a file

PR #2321, opened 2017-05-22
Issue2316 : IDE plugin modifications

skarsaune:
Some changes to the IDE plugin based on observations in #2316
1. Prefer OpenFileHttpService 
2. Do only one call from front end to back end , to have less coupling and allow the backend / IDE to decide optimal resolution of resources
3. Remove some functionality that seems never completed
4. Split out some code from IdeFacade

Can now be tested from hawtio-web : mvn exec:java in both JUnit and Diagnostic/Heap by selecting classes / stack frames and IDEA open with the source code of the files (for instance *.hawt.* classes)
image
image
image

One disadvantage is that IDEA does not communicate back whether the operation succeeded, so it is best effort based.

Have also left the existing used JMX operation, in case an older hawt.io frontend runs against a new hawt.io front end. Is that relevant at all ?

Questions / further steps:
- single out IDEA specific code, to allow support for other IDEs ?
- select preferred IDE from frontend ?
- implement an Eclipse plugin with similar interface to IDEA ?

But before work is done, it is a good time to consider the overall usefulness of the plugin. 
For my own workflow I would typically copy class name to the IDE and go straight to the search function, or copy stack trace / heap dump into "analyse stack trace" / stack trace console . This manual process gives full control (no surprises) and has less technical limitations such as hawt.io and IDEA to run on the same host and that source file must be on disk (e.g not in rt.jar).  Any thoughts @jstrachan , @tadayosi , @davsclaus  ?


davsclaus:
Yeah if its possible to integrate with other IDEs then that would be nice. 

I don't think the older hawtio frontend vs backend is so relevant. The frontend is a html app that is serviced from the same backend installation of hawtio.

I am not sure how many people are using it, but it looks nice for demos, that you can click a link and IDEA shows the source code. But as you said its likely better to just have a copy button, and then people can do as you describe instead. Also you may have a project you work on that you load in IDEA that has all the dependencies setup.

skarsaune:
Then I will remove the old operation that is no longer used. I think I will leave support for other IDEs to a future PR if I have the time. Would someone like to give the current fix a test run? Since this basically is integration on the local machine, it would be nice to test some different setups. 
I myself use a macbook and run IDEA 2017.1 Ultimate. 

Test suggestions as described above:

"tested from hawtio-web : mvn exec:java in either JUnit and Diagnostic/Heap by selecting classes / stack frames and IDEA open with the source code of the files (for instance .hawt. classes)"

skarsaune:
Removed unused methods in backend. No further changes planned, but good if someone else have the chance to test the fix.

oscerd:
@davsclaus I'm gonna merge this one too

oscerd:
Thanks @skarsaune

PR #2341, opened 2017-06-27
ENTESB-6801 - Fix some js memory leak

paoloantinori:


tadayosi:
Hi @paoloantinori, please commit this to master instead. We don't generally maintain 1.4.x unless there's any imminent reason to do so.

paoloantinori:
Hi @tadayosi this is required for JBoss Fuse 6.x that's is currently bound to Upstream 1.4 branch.

tadayosi:
@paoloantinori AFAIK no one actually uses (maintains) 1.4.x branch upstream and master (1.5.x) branch is in sync with JBoss Fuse 6.3.x branch. So it doesn't make sense to maintain 1.4.x for Fuse 6.x.

tadayosi:
FYI, `master` branch is for current 1.5.x versions and we use `2.x` for hawtio v2 (and Fuse 7).

paoloantinori:
> @paoloantinori AFAIK no one actually uses (maintains) 1.4.x branch upstream any more and master (1.5.x) branch is in sync with JBoss Fuse 6.3.x branch. So it doesn't make sense to maintain 1.4.x for Fuse 6.x.

What makes say that?

If I compare current 6.3.x branch with upstream/master I can count 1359 commits that are present only upstream:

```
$ git log --oneline  upstream/master ^origin/1.4.0.redhat-6-3-x | wc -l
1359
```

tadayosi:
Yeah, there are always changes that are only in upstream and not in product branches, like new features and bug fixes contributed from community. But I also believe `master` branch has (and should have) all the fixes from product as well.

I think we prefer Upstream-First approach just because we would like to make sure we won't have regressions in the future, so at least it doesn't make sense to me to commit fixes from product to a community branch no one uses. Yeah, in this light ideally we should first commit fixes from product to hawtio v2 repositories to achieve the goal of Upstream-First. (In this case though it shouldn't be necessary because we don't use ZeroClipboard.js anymore in hawtio v2.)

tadayosi:
I always try to commit fixes from product to `master` (1.5.x) branch because 1) hawtio v1 community is still vibrant there are not a small number of users of v1 and 2) if there are fixes available in v1 master branch and some of hawtio v2 devs finds it useful for v2 then they will eventually port it to v2.

PR #2346, opened 2017-06-29
Document diagnostics

skarsaune:
Have verified in jetty that pages render OK. 
https://github.com/skarsaune/hawtio/blob/document_diagnostics/docs/Plugins.md

Not able to verify link before it is served as part of hawtio github, but follows pattern from other plugins.


tadayosi:
Thanks for the PR.

Issue #2350, opened 2017-07-04
Java Flight Recorder tab should not appear for OpenJDK


skarsaune:
When jfr* operations are not available on com.sun.management:type=DiagnosticCommand (in OpenJDK), then  the Flight Recorder tab should not be shown. 
Will have to look into the retrospective capabilities as used in the JMX plugin.

PR #2351, opened 2017-07-04
#2350 detect what parts of the diagnostic tools are in fact available

skarsaune:
Tested on zulu JDK where Heap tab is shown and selected , and flight recorder tab is hidden. 

All three tabs still appear in Oracle JDK8.

Since the tab breadcrumbs are built in diagnosticsHelpers.ts one alternative would be to only add the breadcrumbs that are valid and pick the URL of the first breadcrumb if any, instead of adding all with isValid filters.


tadayosi:
LGTM. Thanks for the PR!

PR #2352, opened 2017-07-09
Detect more details about the JVM of the discovered agents

skarsaune:
Idea: provide more details about the Java application of the discovered agents. This can be useful if there are a lot of different Java processes running and one wants an idea of which is which before actually connecting. It will only be attempted for connections without authentication (did not find a good way to deal with authentication). A downside is that there may be up to a couple of extra requests for each discovered agent. However the timeout for the discovery process is so long that the user will probably not notice much difference.

On a best effort basis, the following is attempted for each generic detected vm: 
- Main class / jar file and arguments
- Start time of the application (you often have an idea about when a version was deployed)

image


davsclaus:
This looks good, should we get it merged, or do you want to do it yourself, you are on the hawtio team know, welcome onboard, and great work you are doing - awesome

skarsaune:
Thanks! Yeah, I noticed the big bad merge button right there in my sight, but was not sure about the preferred process ..... 
I'm happy to do it once it has passed review. 
Think I will.

Issue #2360, opened 2017-07-24
My observation from getting hawtio running on Heroku


fharms:
Hi

I just want to share some of my observations I have made in my work to get Hawtio working on Heroku instance.

1. First of all it has not been possible for me to get the 1.5.2 WAR deployment working. It return a empty page, but I can see from the page source it retrieving the page. The standalone jar file is working fine. I had to go back to 1.5.0 to find a working WAR file
console error for 1.5.2:
jquery.js:8416 GET http://public:biscuit@localhost:5000/jolokia 403 ()
loggingInit.js:57 [Window] Uncaught SyntaxError: Unexpected token < in JSON at position 0  (<a href="http://localhost:5000/:1">http://localhost:5000/:1:1</a>)
2. Enable security on hawtio was easy, but getting it to work on Heroku Tomcat Webapp-runner was a hassle. The problem was the TomcatUserDatabaseLoginContext expect the tomcat-user.xml is located in tomcat’s conf directory. That’s not the case for Heroku webrunner, in my case it was located in the Dyno App root directory. I got around it by hacking the TomcatUserDatabaseLoginContext and allow to setup a property variable “hawtio.tomcatUserFileLocation” pointing to my tomcat-user.xml user location
3. Connect to a remote jolokia agent also required a bit of work. Since you're only able to access the Heroku dyno from port 80 or 443 I had to configure the jolokia agent to use the servlet agent instead with basic authentication. I got this working with the standalone 1.5.2 jar file, but with the 1.5.0 war file I’m not able to connect to the remote agent. It’s does not report any errors and I can not see any errors in the log file either.

Bottom line I have a hacked version of Hawtio 1.5.0 working on my Heroku instance, but not able to remote connect to a jolokia agent. 

I have created a pull-request and hope you will consider adding my hack to the code, and I hope somebody can help me getting the 1.5.2 war file working.

fharms:
Finally I figure it out. I was missing the "--path /hawtio" to the webrunner. 
```
java $JAVA_OPTS -jar target/dependency/webapp-runner.jar  --path /hawtio --tomcat-users-location ../../tomcat-users.xml --port $PORT target/dependency/hawtio.war
```

oscerd:
@fharms what do you think about documenting this in the hawtio docs? I guess it can be useful.

NKame:
Standalone WAR still not working with 1.5.3: 
Vivaldi browser (based on Blink)
[...]

PR #2361, opened 2017-07-24
Override tomcat user xml location

fharms:
Please see the #2360 issue for a more detailed description on why I have created this pull-request


davsclaus:
Thanks do you mind formatting the source code a bit so its standard, eg spaces around `==` etc.

fharms:
@davsclaus I formatted the code in IDEA, but is there a style I need to follow?

davsclaus:
No there is no checkstyle but the its good to keep the code in similar style

PR #2375, opened 2017-08-16
Fixes #2374 Incorrect header in CORSFilter

### LINKED ISSUE #2374: Incorrect header in CORSFilter
The CORSFilter is writing to a header called 'Access-Control-Allow-Header', which is incorrect.  It should be 'Access-Control-Allow-Headers'.

See [CORS spec](https://www.w3.org/TR/cors/#access-control-allow-headers-response-header)

seryckd-jda:


tadayosi:
Good catch!  Thanks for the pull req.

Issue #375, opened 2013-06-21
Support importing FMC .zip profile/version export format


gashcrumb:
Folks moving from FMC for controlling their fabric will need an easy way of importing their existing profiles, ideally we should be able to handle importing FMC's .zip file export format.

PR #2378, opened 2017-09-04
ENTESB-6558 - Implement a filter to prevent arbitrary file upload to hawtio

hoomanb1:
Change: Added a filter to check uploaded file's signatures (magic numbers) to ensure that the file is permitted to be upload. This allows users to configure/whitelist expected file types as well as max file size as a systemProperty. The systemProperty format is similar to below and can have multiple config separated via semicolon:
``` hawtio.config.fileUpload", "signature=7573746172,offset=257,maxSize=10kb;...``` 
where signature is the file's magic number, offset is a pointer at which magic bytes start and maxSize is the maximum allowed file size. All file type signature information can be found from [Signature Database](https://www.filesignatures.net/index.php?page=all).

### CODE REVIEW

File: hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static java.lang.StrictMath.toIntExact;
+
+public class FileUploadFilter {
+
+    private static final transient Logger LOG = LoggerFactory.getLogger(FileUploadFilter.class);
+    public static final String FILE_UPLOAD_PROPNAME = "hawtio.config.fileUpload";
```
tadayosi: Looks like `hawtio.config.` is for hawtio-config parameters.
http://hawt.io/configuration/index.html
Probably we should use `hawtio.upload.filter` or something?

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
```diff
@@ -0,0 +1,171 @@
...
+        byte[] data = new byte[len / 2];
+        for (int i = 0; i < len; i += 2) {
+            data[i / 2] = (byte) ((Character.digit(s.charAt(i), 16) << 4)
+                + Character.digit(s.charAt(i + 1), 16));
+        }
+        return data;
+    }
+
+    // Exposed for testing
+    public long translateFileSize(String size) {
```
tadayosi: For testing we can just make it package private.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
```diff
@@ -0,0 +1,171 @@
...
+
+import static java.lang.StrictMath.toIntExact;
+
+public class FileUploadFilter {
+
+    private static final transient Logger LOG = LoggerFactory.getLogger(FileUploadFilter.class);
+    public static final String FILE_UPLOAD_PROPNAME = "hawtio.config.fileUpload";
+    private List<MagicNumberFileFilter> filters;
+
+    private void addFilters(MagicNumberFileFilter config) {
```
tadayosi: Why is it private when the `getFilters` method is public?  Should users use `getFilters` to add a custom filter?

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
```diff
@@ -0,0 +1,171 @@
...
+    public List<MagicNumberFileFilter> getFilters() {
+        if (this.filters == null) {
+            this.filters = new ArrayList<>();
+        }
+
+        return filters;
+    }
+
+
+    static final class MagicNumberFileFilter {
```
tadayosi: Why is it package private?  The `getFilters` method being public, doesn't it cause any problems when an user tries to add a custom filter from outside of the package?

hawtio-system/src/test/java/io/hawt/web/FileUploadFilterTest.java
```diff
@@ -0,0 +1,111 @@
+package io.hawt.web;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
+
+import java.io.*;
```
tadayosi: Please don't use `*`.

hawtio-system/src/test/java/io/hawt/web/FileUploadFilterTest.java
```diff
@@ -0,0 +1,111 @@
+package io.hawt.web;
+
+import org.apache.commons.io.FileUtils;
```
tadayosi: Please follow the project's preferred import order.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
```diff
@@ -0,0 +1,171 @@
...
+            this.filters = new ArrayList<>();
+        }
+
+        filters.add(config);
+    }
+
+    public FileUploadFilter() {
+
+        String config = System.getProperty(FILE_UPLOAD_PROPNAME);
+        if (config != null) {
```
tadayosi: No default config?  Can any file be uploaded by default?

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
```diff
@@ -143,4 +147,8 @@ public void update(long pBytesRead, long pContentLength, int pItems) {
         return uploadedFiles;
     }
 
+    private boolean fileAllowed(FileItem fileItem) throws IOException {
+        FileUploadFilter uploadFilter = new FileUploadFilter();
+        return uploadFilter.accept(IOUtils.toByteArray(fileItem.getInputStream()));
```
tadayosi: It seems that the whole content of an uploaded file is loaded to memory. What if the size of a file is huge (e.g. 1GB)?  Doesn't that cause OOM error or performance degradation?

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
```diff
@@ -1,26 +1,25 @@
...
-
 import io.hawt.jmx.UploadManager;
 import io.hawt.util.Strings;
 import org.apache.commons.fileupload.FileItem;
 import org.apache.commons.fileupload.FileUploadException;
 import org.apache.commons.fileupload.ProgressListener;
 import org.apache.commons.fileupload.disk.DiskFileItemFactory;
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.filefilter.MagicNumberFileFilter;
```
tadayosi: Why is it needed when you have your own `MagicNumberFileFilter` class?  If it doesn't compile without this import then it smells like a bug in the code.

tadayosi:
Hi @hoomanb1,

There is one more thing I'm concerned about this fix. Looking at #375, the upload servlet was introduced mainly for uploading `.zip` artifacts for Fabric users. Furthermore, `GitServlet` extends the `UploadServlet`. So we need to make sure applying the file filtering to the upload servlet doesn't break the use cases.

Honestly I'm not sure how `GitServlet` and `UploadServlet` are used in hawtio nowadays. We could possibly just remove them from the community version of hawtio to secure it. But still we need to be conservative on changes to those servlets for JBoss Fuse.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: The idea wasn't meant to allow users to create custom filters programmatically rather allowing a singleton way of accessing the filters, this is primarily for `FileUploadFilter` itself to make these filters available for `accept` method, so as long as users have their filters configured via system property then it should just work.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: With the current solution on detecting via magic numbers it would be a little tricky to blacklist some files out of the box as some file signatures might be different for the same file type, unless we add all possible scenario which could essentially clutter our code base, so I decided to delegate this to users themselves on what content type they really expect to accept via file upload so they can configure this that way.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: True, will fix it.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: They are not suppose to, but you're right filter list should be immutable list.

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
```diff
@@ -1,26 +1,25 @@
 package io.hawt.web;
 
-import java.io.File;
```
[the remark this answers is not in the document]
hoomanb1: Sure

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
```diff
@@ -1,26 +1,25 @@
...
+import org.apache.commons.io.filefilter.MagicNumberFileFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.*;
```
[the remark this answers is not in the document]
hoomanb1: Sure

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
tadayosi: Ah yes making it immutable is also a good practice. What I meant though is that if `MagicNumberFileFilter` is package private, the class should not be visible from external packages. So I wonder what would happen when an user code tries to invoke the `getFilters` method from, say, `com.example.app`?

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
hoomanb1: Fixed

hawtio-system/src/test/java/io/hawt/web/FileUploadFilterTest.java
hoomanb1: Yep

hawtio-system/src/test/java/io/hawt/web/FileUploadFilterTest.java
hoomanb1: Yep

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
hoomanb1: Corrected.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
```diff
@@ -0,0 +1,171 @@
+package io.hawt.web;
+
+import org.apache.commons.io.FileUtils;
```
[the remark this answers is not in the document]
hoomanb1: ok

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: Alright changed to `hawtio.upload.filter`

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: They are not suppose to, I've changed it to `private` and made the list immutable. Essentially this filter is a global filter that only allows user to access the global file filter by invoking `FileUploadFilter fileUploadFilter = FileUploadFilter.newFileUploadFilter()` which gives them access to the global filter by invoking the `getFilters` which is an immutable list of filters. This can be used throughout the application to filter any file upload.

hoomanb1:
@tadayosi , yes agreed `GitServlet` is kind of tightly uses the `UploadServlet`. I'm not sure either as to whether these two are being used at the moment. We can add more to the filter to accommodate for `GitServlet` as well but again this might be extra work if we're not using them. I've had a quick look at fuse/fabric hawtio and can't quite work out if we can export fabric profiles as a zip file at all!

tadayosi:
What's the current status of this?  If you've already fixed them, then please submit another PR. Thanks!

hoomanb1:
Closing, please review #2388 instead.

PR #2387, opened 2017-09-19
ENTESB-6558-V2

hoomanb1:
- Added more logic to inspect file content based on system properties exception configuration: this should allow users to add character exception for ascii file upload. 
- Added logic to inspect zip file content: this ensure zip content complies to file upload restriction e.g. number of file within the zip file
- Added `GitServlet` filter as well: this should ensure that `GitServlet` has it's own filter config out of the box


hoomanb1:
@tadayosi, sorry for the mess on the commit history, please review `a) added content detection logic b) added archive ...` only.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
```diff
@@ -0,0 +1,5 @@
+package io.hawt.web;
+
+public interface FileUploadFilter {
```
tadayosi: Why is it needed?

hawtio-system/src/main/java/io/hawt/web/GitFileUploadFilter.java
```diff
@@ -0,0 +1,40 @@
+package io.hawt.web;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GitFileUploadFilter {
```
tadayosi: Why do we need a whole new class just to configure it with `GIT_CONFIG`?  If there's any awkwardness that requires it then I guess `GlobalFileUploadFilter` should be refactored to make it more flexible.

hawtio-system/src/main/java/io/hawt/web/GlobalFileUploadFilter.java
```diff
@@ -0,0 +1,420 @@
...
+        for (String s: finalProhibitedList) {
+            if (builder.toString().startsWith(s) || builder.toString().contains(s)) {
+                dangerousContentDetected = true; // script file detected
+            }
+        }
+
+        return dangerousContentDetected;
+    }
+
+    /*Getting the maximum file size allowed from the config*/
```
tadayosi: Javadoc style is better.

hawtio-system/src/main/java/io/hawt/web/GlobalFileUploadFilter.java
```diff
@@ -0,0 +1,420 @@
...
+        } finally {
+            zipInputStream.close();
+        }
+
+        return safe;
+    }
+
+    static Set<String> getFinalProhibitedList(List<MagicNumberFileFilter> fileFilter) {
+        Set<String> finalProhibitedList = new HashSet<>();
+        /**
```
tadayosi: Javadoc shouldn't be used here.

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
```diff
@@ -143,4 +160,7 @@ public void update(long pBytesRead, long pContentLength, int pItems) {
         return uploadedFiles;
     }
 
+    private boolean fileAllowed(FileItem fileItem, List<GlobalFileUploadFilter.MagicNumberFileFilter> filters) throws IOException {
+        return GlobalFileUploadFilter.accept(IOUtils.toByteArray(fileItem.getInputStream()), filters);
```
tadayosi: I don't think this review comment of mine in the previous pull req #2378 is addressed correctly yet.
> It seems that the whole content of an uploaded file is loaded to memory. What if the size of a file is huge (e.g. 1GB)? Doesn't that cause OOM error or performance degradation?

Actually, the biggest obstacle that makes me hesitant to merge this pull req is here in two parts:

1. Can it be a new source of vulnerability if an attacker exploits this part to execute a DOS attack?
2. Even if it's safe against DOS, it should still introduce a certain extra load to every request to those servlets even though the default 'Allow All' settings.

Is it the only way to check the content of an uploaded file?

hoomanb1: 1) The underlying upload implementation is via Commons file upload which uses streams file content bearing the `fileSizeMax` that's been set in the `UploadServlet`: 
```
upload.setFileSizeMax(GlobalFileUploadFilter.getMaxFileSizeAllowed(filters));
```
and the file size comes from users config, so this should prevent dos all together. 
2) Well I guess this probably relates to how the underlying app server is configured wrt thread etc.  So I guess loads are expected for the servlets regardless and since we have the content of the file as byte streams then this could be ideal in allocating byte size to ensure we've introduced a threshold for bytes. Finally I don't believe these servlets are going to be under heavy load, do they?

hawtio-system/src/main/java/io/hawt/web/GlobalFileUploadFilter.java
hoomanb1: Will fix this.

hawtio-system/src/main/java/io/hawt/web/GlobalFileUploadFilter.java
hoomanb1: Will fix it

hawtio-system/src/main/java/io/hawt/web/GitFileUploadFilter.java
hoomanb1: Well we could do that but the idea is to allow process wide filter configuration as well as the `GlobalFileUploadFilter`, in this case `GitFileUploadFilter`.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: This is just to introduce a `type` for the implementation instance, in this case `GlobalFileUploadFilter`, this is to ensure if people wants to introduce different filter/behavior then they stick to `FileUploadFilter` category.

hoomanb1:
Closing this one, added #2388 instead.

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
tadayosi: IMO so long as the interface doesn't have any methods, it's not useful. If your intention is to provide users an extension point for behaviours then such a marker interface (= an interface without behaviours) are somewhat odd. If you just want to make a category for filters then just digging a `uploadfilter` package is enough.

hawtio-system/src/main/java/io/hawt/web/GitFileUploadFilter.java
tadayosi: It doesn't make sense to me. In object-oriented programming, you don't need a class to have different configurations for different processes; you just need different instances from a single class. In this case, `GlobalFileUploadFilter`'s `filters` is an instance field so there's no reason you cannot use it for multiple processes with the different instances of it.

`GIT_CONFIG` can be located where a `GlobalFileUploadFilter` instance for `GitServlet` is created, which is `GitServlet` itself.

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
tadayosi: My point is this part:
```
IOUtils.toByteArray(fileItem.getInputStream())
```
which dumps all the content from the input stream to memory. An input stream normally read each chunk of bytes from the request and write it down to a target file in each step, so there should be no performance penalties. But the code dumps all of the bytes into memory at once, which I don't know is really necessary.

And even though you can limit the max size of each uploaded file, they could be cumulatively huge (e.g. 10MB * 100 = 1GB).

Finally, you can just disable the filter altogether when there is no `hawtio.upload.config` system property to avoid the default extra load.

hawtio-system/src/main/java/io/hawt/web/GitFileUploadFilter.java
hoomanb1: Fixed in #2388

hawtio-system/src/main/java/io/hawt/web/FileUploadFilter.java
hoomanb1: OK will remove it.

hawtio-system/src/main/java/io/hawt/web/UploadServlet.java
hoomanb1: So for this:
>Finally, you can just disable the filter altogether when there is no hawtio.upload.config system property to avoid the default extra load.
It already does that see `UploadServlet`:
```
GlobalFileUploadFilter globalFilter = GlobalFileUploadFilter.newFileUploadFilter();
        // By default file upload is not permitted unless it is configured
        if (globalFilter.getFilterConfig().isEmpty()) {
            response.sendError(HttpServletResponse.SC_FORBIDDEN, "File upload is rejected for security reasons");
        }
```
Regarding your point on `IOUtils.toByteArray(fileItem.getInputStream())`, got your point, the stream should be closed.

PR #2393, opened 2017-10-10
JFR support for Java 9

skarsaune:
Background: The DiagnosticCommand operations have both differing input requirements and response between Java 8 and Java 9.
Goal: Adjust the implementation so that it supports both versions.
Implementation details: 

- Drop compress option as it is no longer supported
- Base operations for existing recording on recording name rather than number
- Allow starting new recordings even if there are stopped ones (are not removed by stop in Java 9)
- Support maxsize limit of recordings

Tested successfully in Java versions 9+181, 1.8.102 & 1.8.152.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ hawtio-config
 node/
 bower_components/
 .vscode
+*.jfr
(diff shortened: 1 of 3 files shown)


skarsaune:
@davsclaus @tadayosi : Regarding desire to release now. I just found a remaining small issue here. 
It only occurs in combination of running time limited flight recordings on Java 9. 
It would be great if I had a chance this evening to see if I can manage to find a solution before we merge and release. 
I can give updated status by tomorrow.

davsclaus:
@skarsaune yeah sure no rush we can do a hawtio release in a few days if so

skarsaune:
Ok, I am satisfied with the latest changes. Will merge this evening unless anyone has improvement suggestions.

PR #2419, opened 2017-12-12
Diagnostics cleanup and tests

skarsaune:
* Some cleanup learned while porting to hawtio-jmx (remove var, dead code, module -> namespace, , more strong typing etc.)
 * Make non trivial logic testable and write regression tests


oscerd:
Thanks

PR #2438, opened 2018-02-17
Spring: add ability to disable Hawtio via properties

wigbam:
Sometimes it might not be desirable to start Hawtio automatically as soon as it is included on the class path. This enhancement allows to disable Hawtio via configuration property "hawtio.enabled" which defaults to "true", but can be overridden if necessary.


tadayosi:
Looks good to me.

PR #2442, opened 2018-03-03
improved SpringBoot integration

### LINKED ISSUE #2138: Hawt.io embedded in spring boot app cannot use different web context

wigbam:
The purpose of this commit is to improve integration with SpringBoot. It introduces the following changes:

1. fixes #2138
    hawtio now respects `"server.contextPath"`, `"server.servlet-path"` and `"management.context-path"` properties and it should function no matter how those paths are configured.
    
    hawtio also works correctly regardless of whether `"management.port"` is the same as `"server.port"` or is different.

2. Spring properties are integrated into hawtio
    hawtio properties are now sourced from Spring environment (e.g. `"hawtio.*"` properties declared in `"application.properties"` file) and passed to `ConfigManager`.
   
    Unfortunately, some servlets/filters ignore `ConfigManager` and read system properties directly (e.g. `SessionExpiryFilter` reads `"hawtio.noCredentials401"` property directly) - some further work is needed to fully integrate such entities.

3. hawtio endpoint can be configured using canonical actuator properties
    hawtio now respects `"endpoints.hawtio.enabled"`, `"endpoints.hawtio.path"` and `"endpoints.hawtio.sensitive"` properties

4. hawtio can be run from the root of the management context
    in certan scenarios where hawtio is the only management endpoint exposed it might be desirable to run it from the context root, e.g. `http://localhost:10001/`

5. hawtio respects Jolokia endpoint path
    SpringBoot allows to override default Jolokia endpoint path by setting `"endpoints.jolokia.path"` property. Hawtio now respects the actual Jolokia endpoint path instead of using the hard-coded `"/jolokia"` value.

I have added unit tests for all new and modified classes as well as MVC integration tests for all conceivable scenarios. I have also tested it manually both in stand-alone mode and as a SpringBoot application with various port/path combinations.

## Note on SessionExpiryFilter

While writing tests for SessionExpiryFilter it became clear that this fellow does significantly more work that the name would suggest, namely:

1) handles `"/refresh"` requests and retuns `"ok"`
2) handles authentication checks before allowing requests to certain sub-contexts
3) handles Jolokia 401 logic
4) finally it does actually update session access time

Looks like `SessionExpiryFilter` is in needs of some refactoring to move out all of the logic not related to session expiry into appropriate exisiting/new components?


davsclaus:
Thanks for all this hard work - keep em coming.

I think it would be good if @tadayosi also took a look before its merged.

tadayosi:
I've been busy with other tasks right now. Let me revisit it in a couple of days.

tadayosi:
@wigbam Thanks. Awesome work!

Here are a couple of small (trivial) feedback before merging it.

1. What does `endpoints.hawtio.sensitive` mean?  What happens when it's turned `true`?

2. Maybe we can move `hawtio-springboot/src/integration-test/` to `src/test` and just ignore tests with `*IT.java` in the normal `mvn test` runs. We can then run itests using `integration` profile (i.e. `mvn test -Pintegration`).

wigbam:
1) `endpoints.hawtio.sensitive` is a flag which tells Spring whether unauthenticated access the specific actuator endpoint should be allowed or not. Unauthenticated in this context relates to Spring's own authentication infrastructure, not hawtio's. In Spring 2 this property was dropped altogether.

    After reviewing and more testing I have to be corrected - support for `endpoints.hawtio.sensitive` is only partial at this point, since Spring is not clever enough to intercept all requests coming under hawtio path (e.g. it allows all static class path resources registered using `addResourceHandlers()` without a problem)
But this was actually my next project - integrating Spring Security with hawtio, so I will have a look at it in the future.

2) you mean you would like to run integration tests only if `itests` profile is activated? I can certainly amend the `pom.xml` so that `src/integration-test/` directory is added only under `itests` profile. Generally, a common approach is to keep unit and integration tests in separate folders so that they don't get mixed up. This way, for example it becomes trivial to run unit tests from IDE without triggering integration tests in the process. But, your call ultimately.

tadayosi:
@wigbam OK, regarding the 2nd point I was just concerned that `src/integration-test` is not a standard mvn directory. I understand that it's easy to run itests when you're on IDE, so let's leave it as-is for now.

In the future, we should create `itests` directory and put more itests in it. Then we could move the spring-boot itests under that directory.

PR #2456, opened 2018-03-31
Use relative paths for icon images

gzurowski:
This commit updates all icon image references to use relative paths. This makes sure that icon images can be properly requested even when running hawtio with a custom context path.

This issue was found when running Spring Boot and hawtio with a custom context path. All images in the "Route Diagram" tab were not rendered properly, because they were requested using absolute paths starting with `/img/`.


tadayosi:
@gzurowski Thank you!  This looks good. Merging.

PR #2458, opened 2018-04-02
Use Camel version property from parent

gzurowski:
The Spring Boot sample should use the same Camel version as configured in the parent.


tadayosi:
Thanks!

PR #2459, opened 2018-04-02
Improve README for the Spring Boot sample

gzurowski:
Include details about configuring alternative context paths and provide
sample URLs.


tadayosi:
Awesome!

PR #2481, opened 2018-06-29
2.0.x -- allow using SSL by specifying a keyStore and keyStorePass form embedded hawtio

myrlyn:
SSL is already available via container configuration for Hawtio running in external Servlet Containers (such as tomcat, or a managed Jetty instance), but not for embedded use.  

This patch adds support for SSL using embedded Jetty. 

I needed this for one of my own projects, and decided to offer this upstream. 

So far the only remaining issue is that the redirect from "https://host:port/hawtio" redirects to "http://host:port/hawtio/" instead of "https://host:port/hawtio".  

Other redirects (such as the ones to login, if authentication is enabled) work properly.

### DIFF
--- a/hawtio-embedded/pom.xml
+++ b/hawtio-embedded/pom.xml
@@ -17,6 +17,11 @@
       <artifactId>jetty-webapp</artifactId>
       <version>${jetty-version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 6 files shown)


tadayosi:
@myrlyn Thank you for your great contribution!

A couple of minor change requests before merging your work.

* Squash the commits into one commit
* Use spaces instead of tabs for indentation

Thanks!

myrlyn:
Will do!

PR #2488, opened 2018-07-11
enable configuration for SSL for jetty

myrlyn:
Squashed merges, converted tabs to spaces.

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/auth/LogoutServlet.java
+++ b/hawtio-system/src/main/java/io/hawt/web/auth/LogoutServlet.java
@@ -40,7 +40,13 @@ public class LogoutServlet extends HttpServlet {
             session.invalidate();
         }
         request.logout();
-        response.sendRedirect(request.getContextPath() + AuthenticationConfiguration.LOGIN_URL);
+        String schem=request.getServletContext().getInitParameter("scheme");
+        if (null == schem) {
+            schem = "http";
+        }
+        System.out.println(schem+" --logoutservlet");
+        String redir = schem+request.getServerName()+":"+request.getServerPort()+request.getContextPath()+AuthenticationConfiguration.LOGIN_URL;
+        response.sendRedirect(redir);
     }
 
 }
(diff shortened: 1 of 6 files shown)


tadayosi:
Thank you for your contribution!

hawtio-system/src/main/java/io/hawt/web/auth/LogoutServlet.java
```diff
@@ -40,7 +40,13 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
             session.invalidate();
         }
         request.logout();
-        response.sendRedirect(request.getContextPath() + AuthenticationConfiguration.LOGIN_URL);
+        String schem=request.getServletContext().getInitParameter("scheme");
```
alexkieling: @myrlyn @tadayosi Why do we need this change? Also, when the `scheme` init parameter isn't set and we are using https, the code redirects to http. Isn't that wrong?

tadayosi: @alexkieling It's because of #2481

jamesnetherton: The redirect logic is definitely wrong for non-embedded scenarios IMO. I did a quick test on WildFly with HTTPS which confirms this. Notice the location header in the response.
```
$ curl -v -L -k https://localhost:8443/hawtio

GET /hawtio HTTP/1.1
Host: localhost:8443
User-Agent: curl/7.52.1
Accept: */*
location: https://localhost:8443/hawtio/
content-length: 0
date: Tue, 18 Sep 2018 06:23:43 GMT

GET /hawtio/ HTTP/1.1
Host: localhost:8443
User-Agent: curl/7.52.1
Accept: */*
HTTP/2 302 
access-control-allow-origin: *
x-xss-protection: 1
x-frame-options: SAMEORIGIN
location: http://localhost:8443/hawtio/auth/login
content-length: 0
date: Tue, 18 Sep 2018 06:23:43 GMT
```

alexkieling: @jamesnetherton could you please create a new GitHub issue for this bug?

PR #2517, opened 2018-09-21
fix(spring-boot): Spring Boot custom context paths do not work

jamesnetherton:


hawtio-system/src/main/java/io/hawt/web/filters/BaseTagHrefFilter.java
```diff
@@ -0,0 +1,137 @@
...
+    private String basePath;
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        final String path = filterConfig.getInitParameter(PARAM_APPLICATION_CONTEXT_PATH);
+        applicationContextPath = path != null ? path : "";
+
+        final ServletContext context = filterConfig.getServletContext();
+        final String contextPath = context.getContextPath();
+        if (contextPath == null || contextPath.isEmpty()) {
```
alexkieling: I think contextPath will never be null.

Issue #2515, opened 2018-09-18
hawt.io/jolokia gives user blank screen if JMX call throws an NPE


exabrial:
Hawtio doesn't work out of the box.

1. Drop an unsecured jolokia agent on a remote machine
1. Down the hawtio 2.10 jar. 
1. Run it from the command line
1. Try to connect to remote jolokia agent. You'll be prompted for a password (even through there is none)

tadayosi:
Have you configured `hawtio.proxyWhitelist`?  Also, have you checked the CORS settings for the remote jolokia agent?

exabrial:
Thanks for the reply! So we configured the whitelist, but now we're getting a white screen. Does the proxy do CORS preflights? I wouldn't think that configuring it would be necessary.

exabrial:
Also it looks like it got data back from the agent, so it's likely not cors

exabrial:
This took _a lot_ of problem solving but we figured it out! This patch here breaks hawtio: https://github.com/apache/tomee/pull/166

exabrial:
Take a look at that PR, I put some notes on how to reproduce the issue. I can send you the config file that breaks hawtio if I can email you

exabrial:
Actually, I think I know what it is. Trying creating an MBean that throws an `NullPointerException` when it tries to list the attributes/operations

tadayosi:
Great investigation. So can/should TomEE handle the NPE?  Or should Hawtio be robust enough to handle it?

exabrial:
Definitely a bug in both pieces of software that needs to be fixed. I'm not sure on the jmx side of if the problem is jolokia or hawtio though

exabrial:
was able to get you a stack trace by putting a exception breakpoint into the JVM:

```
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sun.reflect.misc.Trampoline.invoke(MethodUtil.java:71)
	at sun.reflect.GeneratedMethodAccessor48.invoke(Unknown Source)
[...]

PR #2519, opened 2018-09-26
fix(hawtio-system): #2515 - hawt.io/jolokia gives user blank screen i…

tadayosi:
…f JMX call throws an NPE


jamesnetherton:
Looks good. It'd be nice if we could add a basic unit test for `RBACRegistry` - would give us something to build on if we hit any future issues with it.

tadayosi:
@jamesnetherton thanks for the feedback. Added a basic unit test class. Note the original issue is a bit hard to test (maybe not worth crafting a complicated mock test for the issue) because the API `MBeanInfo.getAttributes()` should normally not return null.

PR #2522, opened 2018-10-02
fixes #2521

### LINKED ISSUE #2521: Refactor JmxTreeWatcher to remove needless logging
It's 2018. Do we really need hawt.io to log "Don't cha wish your console was hawt like me? ;-)" at startup?

benemon:
Small change to address issue #2521


hawtio-system/src/main/java/io/hawt/jmx/JmxTreeWatcher.java
```diff
@@ -61,7 +61,7 @@ public void init() throws Exception {
             if (Strings.isNotBlank(text)) {
```
alexkieling: I think you can remove this `if` statement.

benemon: To be fair, the whole thing can now probably just be boiled down to `LOG.info("Welcome to hawtio " +  getVersion());`

tadayosi: Use

    LOG.info("Welcome to hawtio {}", getVersion());

instead of

    LOG.info("Welcome to hawtio " + getVersion());

tadayosi:
Also we now follow this commit message guideline: https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines
So should the commit as well. Squashing them is also preferable for such small changes.

benemon:
> Also we now follow this commit message guideline: https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines

Would be good to have that in the Contribution docs.

I've squashed the commits and updated the log message as requested.

jamesnetherton:
With `{}` placeholders we don't need to use string concatenation. So it should look like:

`LOG.info("Welcome to hawtio {}", getVersion());`

If you can make that correction, we can get this merged.

benemon:
Sorry, I misread @tadayosi 's comment as providing a delimiter between the message and the version.

Updated and re-pushed.

PR #2530, opened 2018-10-24
chore: Use spring-boot-starter-undertow instead of jetty

jamesnetherton:
@tadayosi I assume this change is fine? The springboot-authentication quickstart seems to work fine with Undertow, and it brings it into line with the normal springboot quickstart.


jamesnetherton:
Ok, I'll merge as-is. I raised #2531 as a follow up task.

Having the Jetty libs + Undertow didn't seem to have any ill effects when I tested, so it should be ok for now.

PR #2533, opened 2018-10-29
refactor(hawtio-system): Move LoginRedirectFilter unsecuredPaths init parameter to static constant

jamesnetherton:


alexkieling:
Looks good

PR #2535, opened 2018-10-31
feat: Add support for Spring Boot 2

jamesnetherton:
I switched `hawtio-springboot` to use Spring Boot 2.x and created a new `hawtio-springboot-1` which uses the Spring Boot 1.x release line. I think this will make life easier if and when we eventually decide to drop 1.x support (ideally when 1.x is officially EOL). Therefore, 2.x is now considered the 'default' for Hawtio (I mention this in the CHANGELOG).

There's a bit of (inevitable) duplication between the 1.x & 2.x modules, which I plan to address in future.


jamesnetherton:
`http://localhost:10001/actuator/hawtio`

There are some notes in one of the example README files on how to customize that:

https://github.com/jamesnetherton/hawtio/tree/spring-boot-2/examples/springboot

Issue #2462, opened 2018-04-06
Support JDK 11 LTS


tadayosi:
JDK 11 will be released this fall. We should make sure both hawtio v1 and v2 build and run on Java 11 once it's released. I think the biggest task among others is to define the module for hawtio.

We can start this task right now using JDK 10.

ffang:
Hi @tadayosi ,

I have made the master build and pass all tests with OpenJDK 11. I will send a PR soon.

Cheers
Freeman

PR #2538, opened 2018-11-07
#2462 make the build and all tests pass with JDK11

ffang:


tadayosi:
Looks great. Thank you!

ffang:
@tadayosi Thanks!
Btw, I also confirmed the change is OK for JAVA8

tadayosi:
@ffang Yes, I also ran it with Java 8 and saw no problems!

By the way, I rushed too early to release Hawtio 2.3.0, just before having this pull req. Let's release the Java 11 support in 2.4.0!

PR #2572, opened 2019-03-20
Add unit tests for io.hawt.util.Files

JohnLBergqvist:
Hi,

I've analysed your code base and noticed that `io.hawt.util.Files` is not fully tested.

I've written some tests for the functions in this class with the help of [Diffblue](https://www.diffblue.com/) [Cover](https://www.diffblue.com/products).

Hopefully these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.

### DIFF
--- a/hawtio-util/pom.xml
+++ b/hawtio-util/pom.xml
@@ -32,5 +32,23 @@
       <version>${slf4j-version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.20.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


tadayosi:
Thank you for your contribution. We love contributions. However, I ran the tests and they took 8s on my laptop. Considering other util tests just run under 0.2s and the importance of `Files` class in this project, I think it's a bit overload. We'd love if you could contribute a lighter test for `Files` like `HostsTest` and `StringsTest`.

By the way, currently this project rather looks for contributions of e2e tests on containers such as Karaf, WildFly, so on.

@alexkieling @jamesnetherton @astefanutti Feel free to add comments if you have anything to add or think otherwise.

tadayosi:
Let me close it for now as we have no responses since the last comment.

PR #2590, opened 2019-07-11
Fix #2588 - Check for blank string

### LINKED ISSUE #2588: ContentSecurityPolicyFilter not testing for empty string release 2.7.0
String keycloakConfigFile = getConfigParameter(KeycloakServlet.KEYCLOAK_CLIENT_CONFIG);
        if (System.getProperty(KeycloakServlet.HAWTIO_KEYCLOAK_CLIENT_CONFIG) != null) {
            keycloakConfigFile = System.getProperty(KeycloakServlet.HAWTIO_KEYCLOAK_CLIENT_CONFIG);
        }
results in an empty string and not a null when the the parameter is not set or is is set but is blank.
line 46 should rather read 
if (keycloakConfigFile != null && !keycloakConfigFile.trim().isEmpty()) {

exabrial:


alexkieling:
@exabrial, would you be able to contribute a unit test for this bug as well?

exabrial:
A good unit test is pretty hard because of the nest of static methods (instead of injection) but this one proves the NPE doesn't occur anymore

alexkieling:
@exabrial thank you

Issue #2581, opened 2019-04-24
SamplePlugin of hawtio-example-springboot not working


sebthom:
The [hawtio-example-springboot](https://github.com/hawtio/hawtio/blob/master/examples/springboot) defines and configures a sample plugin:

https://github.com/hawtio/hawtio/blob/ddca2b386e277fad15b29cea0c6b8d9af657e199/examples/springboot/src/main/java/io/hawt/example/spring/boot/SampleSpringBootService.java#L20-L24

When launching the application the plugin is neither displayed nor executed.

Atm. I am wondering if the whole HawtPlugin implementation of the hawtio-springboot is working at all with hawt.io 2.x.

alexkieling:
@tadayosi @jamesnetherton I don't Hawtio v2+ supports HawtPlugin. Is that right?

tadayosi:
I have plan to work on it this week.

alexkieling:
Great, Tadayoshi. Contact me if you need any help.

PR #2600, opened 2019-09-26
feat(plugin): restore dynamic custom plugin loading

### LINKED ISSUE #2508: Restore custom plugin feature for v2
This is required for supporting custom plugins for Hawtio v2. Thus it blocks #2477.

This includes:
- Port hawtio-plugin-mbean to v2

tadayosi:
Fixes #2508

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 #### 2.8.0 (To be released)
 
+* Dynamic custom plugin support [#2508](https://github.com/hawtio/hawtio/issues/2508)
+
 #### 2.7.1
 
 * Bug fixes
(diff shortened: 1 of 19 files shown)


tadayosi:
@jamesnetherton Thanks for your review. Yeah I noticed that but then came up to the understanding that Hawtio on Spring Boot might not really need PluginServlet, which is used to dynamically search plugin MBeans and find the plugin js to load so that you don't need to repackage hawtio.war. Instead, what the existing Spring Boot examples suggest is that you can just provide the `/plugin` endpoint yourself in the SB app and `HawtPlugin` class is the helper class for it. Because we repackage an app every time for Spring Boot we can choose a different approach from other heavy-weight containers such as Karaf.

tadayosi:
BTW, I'll work on #2581 soon after this one to make SB custom plugin work as well.

PR #2603, opened 2019-10-25
feat(hawtio-console-assembly): ENTESB-10461 - provide configuration for customization of login page text (banner)

tadayosi:


tadayosi:
@akieling Yes but unzipping and modifying `hawtconfig.json` is the exact approach that we document for customers right now:
https://access.redhat.com/documentation/en-us/red_hat_fuse/7.4/html/managing_fuse/manage-monitor-fuse-karaf#fuse-console-branding-karaf
IMO, it doesn't make sense to provide a different way for customisation of the login page without changing them altogether.

Also the information (description and links) may be too long for passing them as env vars or system properties. That said I agree that the current way is a bit cumbersome and there should be room for improvement.

Why don't we file another enhancement to improve how to customise `hawtconfig.json` altogether and tackle the issue as a separate task?

akieling:
Sounds good to me.

tadayosi:
Thanks. Filed #2604 for the `hawtconfig.json` improvement.

Issue #2604, opened 2019-10-28
Provide a better way for users to modify hawtconfig.json


tadayosi:
See: hawtio/hawtio#2603

Currently users need to unzip `hawtio.war` and pack it again to modify `hawtconfig.json`:
https://access.redhat.com/documentation/en-us/red_hat_fuse/7.4/html/managing_fuse/manage-monitor-fuse-karaf#fuse-console-branding-karaf

There should be a more sophisticated way for users to customise the configurations.

akieling:
@tadayosi I think we could leverage the existing `io.hawt.system.ConfigManager` class to customize `hawtconfig.json`. It supports both System properties and JNDI, right? We could either create a single property pointing to an external version of `hawtconfig.json` or multiple properties customizing specific fields. Customizing images is a bit tricky because the new image would need to be placed inside the hawtio jar/war or served by an external Web server.

tadayosi:
@akieling That's an idea, but `hawtconfig.json` is also read from hawtio-core so we might need to find a way to influence the front-end component hawtio-core from the server-side config properties.

Another option might be to use the custom plugin mechanism:
https://github.com/hawtio/hawtio/blob/master/examples/simple-plugin/src/main/webapp/plugin/simplePlugin.js
and invoke `hawtioPluginLoader.registerPreBootstrapTask()` there to let users directly modify `window['hawtconfig']` programmatically. ... Wait, it sounds like it's already possible?

The Login component doesn't share the same code with hawtio-core `config-loader.ts` so we have to modify it to fully support the way anyway, though.
https://github.com/hawtio/hawtio/blob/hawtio-2.8.0/hawtio-console-assembly/app/src/login/login.component.ts#L21-L29

tadayosi:
Note also that hawtio-online and hawtio-operator depend on `hawtconfig.json` as well. I'm not sure if the current custom plugin mechanism works well with hawtio-online & hawtio-operator, and if there's even need for customising `hawtconfig.json`. But I assume it's just a matter of modifying the `hawtconfig.json` config map so it's already sophisticated enough on OpenShift?  @astefanutti WDYT?

astefanutti:
@tadayosi yes, hawtio-online relies on `hawtconfig.json` as a config map. If `hawtio-operator` is used to deploy Hawtio Online, the operator watches changes to the config map and automatically triggers a rolling update, so that the changes are taken into account automatically.

akieling:
[...]

PR #2608, opened 2019-12-04
feat(plugin): ENTESB-12162 - Provide users with a better way to customise hawtconfig.json

tadayosi:
See https://issues.jboss.org/browse/ENTESB-12162

This opens `/plugin` as an unsecure path so custom plugins can be loaded before authentication. It also adds a plugin example that demonstrates how you can customise `hawtconfig.json` through it.


tadayosi:
Thanks @jamesnetherton @akieling. Just modified commit comment so it also refers to #2604.

PR #2611, opened 2020-01-07
feat(hawtio-system): ENTESB-11696 - support authentication with Spring Security

tadayosi:
This adds Spring Security authentication support as a preliminary step to implement support for Keycloak on Spring Boot.


pom.xml
```diff
@@ -58,7 +58,7 @@
     <build-helper-maven-plugin-version>1.12</build-helper-maven-plugin-version>
 
     <spring-version>5.1.10.RELEASE</spring-version>
-    <spring-boot-version>2.1.9.RELEASE</spring-boot-version>
+    <spring-boot-version>2.2.2.RELEASE</spring-boot-version>
```
jamesnetherton: Maybe this version bump should be a separate commit. We should bump the `spring-version` property too.

Also, I thought that JMX was [disabled by default](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes#jmx-now-disabled-by-default) in Spring Boot 2.2.x, so the example projects will likely need to set `spring.jmx.enabled=true`.

tadayosi: good catch. I'll do that.

tadayosi: For `spring.jmx.enabled=true` let's do that in a commit to upgrade SB to 2.2.x.

PR #2613, opened 2020-01-17
feat(hawtio-console-assembly): ENTESB-12346 - support Keycloak integration on Spring Boot

tadayosi:
This is the final piece of https://issues.redhat.com/browse/ENTESB-12346. Please review it. Thanks.


tadayosi:
@jamesnetherton Nice suggestion. One problem, though, is that it may introduce a dependency to keycloak libs in hawtio-springboot. Probably it's better to create another project, hawtio-springboot-keycloak, for the shared configurations instead of putting them into hawtio-springboot. What do you think?

jamesnetherton:
Good point about the keycloak dependencies. Having a dedicated `hawtio-springboot-keycloak` would be good.

tadayosi:
@jamesnetherton OK, thanks. I'll do the improvement in a next commit.

PR #2616, opened 2020-01-20
feat(hawtio-springboot): ENTESB-12346 - move reusable configs from examples into hawtio-springboot-keycloak

### LINKED ISSUE #2614: Move reusable configurations in springboot-keycloak example into hawtio-springboot-keycloak
See #2613.

> it'd be better to have most of the Keycloak configuration beans as part of hawtio-springboot, conditional on the presence of property `hawtio.keycloakEnabled`. Otherwise everybody has to duplicate the same security configuration bits in their apps.

tadayosi:
Fixes #2614

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -82,6 +82,11 @@
       </dependency>
       <dependency>
         <groupId>io.hawt</groupId>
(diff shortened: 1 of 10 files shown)


tadayosi:
Thanks!

PR #2617, opened 2020-01-23
Support caching of ActiveMQ Artemis Address/Queue MBeanInfos

sebthom:
Address and Queue MBeans have different names in Artemis compared to ActiveMQ 5. 
This PR implements the same MBean caching mechanims for Artemis. Without caching HawtIO UI fails to load for Artemis servers with some hundred or thousand queues.

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/jmx/RBACRegistry.java
+++ b/hawtio-system/src/main/java/io/hawt/jmx/RBACRegistry.java
@@ -250,25 +250,40 @@ public class RBACRegistry implements RBACRegistryMBean {
      */
     private String isSpecialMBean(ObjectName nameObject) {
         String domain = nameObject.getDomain();
-        if ("org.apache.activemq".equals(domain)) {
-            String destinationType = nameObject.getKeyProperty("destinationType");
-            // see: org.apache.activemq.command.ActiveMQDestination.getDestinationTypeAsString()
-            if ("Queue".equals(destinationType)) {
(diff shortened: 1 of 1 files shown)


sebthom:
@jamesnetherton done

jbertram:
@jamesnetherton, both ActiveMQ 5.x and ActiveMQ Artemis use Hawtio 1.5.x, but I see that the 1.x branch has been deprecated as of October 24, 2019. Can you confirm that there will be no further releases of 1.5.x? If so, can you recommend any documentation on how to migrate a plugin from 1.5.x to 2.x?

jamesnetherton:
Yes, we're not actively maintaining 1.x so there are no further releases planned from that branch.

@tadayosi do we have any docs around how to migrate plugins?

tadayosi:
No, we don't. You can just refer to the existing builtin plugins and some custum plugin examples to know what a new plugin for 2.x would look like:
https://github.com/hawtio/hawtio-integration/tree/master/plugins/activemq
https://github.com/hawtio/hawtio/tree/master/examples/simple-plugin

Worth noting that we follow this AngularJS styleguide for 2.x plugins (while it's totally optional to follow this for external plugins):
https://github.com/toddmotto/angularjs-styleguide/tree/master/typescript

tadayosi:
Btw, if that Artemis bug is urgent then we can try releasing a hotfix 1.5.x release for Artemis to earn some time for the project to migrate its plugin to 2.x.

PR #2629, opened 2020-05-04
Specify signature for dumpAllThreads operation

TomyLobo:
In Java 11, there are now two overloads for this operation.
Jolokia will thus require you to specify the overload and throw an exception if you don't.
This patch fixes that.

Whether this issue is still present in 2.x, I don't know.
I'm only using hawtio through ActiveMQ Artemis.
Once this is merged and released, someone with an Apache JIRA account needs to open an issue telling them to upgrade.


tadayosi:
Thank you for your contribution, but we don't maintain 1.x branch any more. AFAIK Artemis is in the process of migrating to Hawtio 2.x too.

Issue #2658, opened 2020-11-05
OWASP Dependency Check Issues


jellisgwn:
The wildfly platform war ends up containing a copy of log4j 1.2.17:

```
$ unzip -t hawtio-wildfly-2.11.1.war | grep log4j-
    testing: WEB-INF/lib/log4j-1.2.17.jar   OK
```

which gets flagged by the OWASP Dependency Check plugin for CVE-2019-17571. 

The implementation of the log4j 1.x API provided by JBoss Logging is not susceptible to this vulnerability, and should make it unnecessary to include this dependency in the Wildfly WAR.

Proposed patch:

```
Index: platforms/wildfly/pom.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- platforms/wildfly/pom.xml	(revision 457c699f0f94066ada1bec60524d070d2df6eb7d)
+++ platforms/wildfly/pom.xml	(date 1604564010733)
@@ -54,7 +54,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-war-plugin</artifactId>
         <configuration>
-          <packagingExcludes>**/lib/slf4j*.jar,**/lib/commons-logging-*.jar,**/classes/log4j.properties,**/lib/netty-*.jar,**/lib/plexus-classworlds-*.jar,**/lib/plexus-component-annotations-*.jar,**/lib/plexus-interpolation-*.jar,**/lib/plexus-utils-*.jar,**/lib/wagon-ahc-*.jar,**/lib/wagon-provider-api-*.jar,**/lib/xbean-reflect-*.jar</packagingExcludes>
+          <packagingExcludes>**/lib/log4j*.jar,**/lib/slf4j*.jar,**/lib/commons-logging-*.jar,**/classes/log4j.properties,**/lib/netty-*.jar,**/lib/plexus-classworlds-*.jar,**/lib/plexus-component-annotations-*.jar,**/lib/plexus-interpolation-*.jar,**/lib/plexus-utils-*.jar,**/lib/wagon-ahc-*.jar,**/lib/wagon-provider-api-*.jar,**/lib/xbean-reflect-*.jar</packagingExcludes>
           <failOnMissingWebXml>false</failOnMissingWebXml>
         </configuration>
       </plugin>
```

PR #2660, opened 2020-11-12
exclude-log4j-from-hawtio-wildfly-war

jellisgwn:
Log4j shows up in QWASP Dependency Check report for CVE-2019-17571.

As Wildfly provides an implementation of log4j 1.x which addresses this problem the jar is not needed in the WAR.


jellisgwn:
Fix for https://github.com/hawtio/hawtio/issues/2658

tadayosi:
Thanks @jellisgwn

PR #2666, opened 2020-12-18
Upgrade Google Guava to 30.1-jre

### LINKED ISSUE #2665: OWASP Dependency Check Issue - guava-27.1-jre
The wildfly platform war ends up containing a copy of guava-27.1-jre:
```
$ unzip -t hawtio-wildfly-2.12.0.war | grep -e 'guava-'
    testing: WEB-INF/lib/guava-27.1-jre.jar   OK
```
which gets flagged by the OWASP Dependency Check plugin for [CVE-2020-8908](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2020-8908)

Propose upgrade POM version to guava-30.1-jre.

jellisgwn:
Fixes #2665

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <commons-logging-version>1.2</commons-logging-version>
     <exec-maven-plugin-version>1.5.0</exec-maven-plugin-version>
     <frontend-maven-plugin-version>1.6</frontend-maven-plugin-version>
-    <guava-version>27.1-jre</guava-version>
+    <guava-version>30.1-jre</guava-version>
     <hamcrest-version>1.3</hamcrest-version>
     <httpclient-version>4.5.13</httpclient-version>
     <!-- make sure to align jackson with fuse/karaf 4.2 -->


oscerd (changes requested):
As I said in the issue, I'd be careful with this.

jellisgwn:
@oscerd thanks for the input. what are you requesting is changed?

oscerd:
Guava is known for introducing breaking changes and incompatibilities on each release. That's why I'm saying to be careful.

Also the CVE is a low impact one and we don't know if hawtio is really vulnerable or not. Static tool like Owasp Dependency Check are just reporting if a library is labelled or not with a CVE.. but there is no evidence about exploitability of the flaw.

jellisgwn:
Unfortunately "be careful" is not actionable. If specific areas of the code are heavily reliant of specific Guava behavior it would possible to validate that they have test coverage or can be manually tested.

For many organizations It is not realistic to ignore CVEs based on lack of direct evidence of an exploit. Hence my attempts to follow up on the reports update to non-vulnerable versions. OWASP Dependency Check is far from perfect, and that's why i'm also creating pull requests for false positives it reports against hawt.io

oscerd:
I understand that for many organizations is not realistic to ignore CVEs, but I still do believe that sometimes it would be useful to check the real impact of a vulnerability. Otherwise the value of static tool like OWASP dependency check will become useless.

jellisgwn:
https://github.com/google/guava

See point 2 in the section **Important Warnings**

> APIs without @Beta will remain binary-compatible for the indefinite future. (Previously, we sometimes removed such APIs after a deprecation period. The last release to remove non-@Beta APIs was Guava 21.0.) Even @Deprecated APIs will remain (again, unless they are @Beta). We have no plans to start removing things again, but officially, we're leaving our options open in case of surprises (like, say, a serious security problem).

Is the project using any @Beta tagged APIs? Otherwise the expectation should be that APIs are binary-compatible. I know that there have been issues in earlier Guava releases, but it looks like they have moved to address that.

tadayosi:
Thanks for your contribution!

PR #2669, opened 2021-02-01
Create synthetic FlightRecorder operation to be downloaded as a stream

skarsaune:
https://github.com/hawtio/hawtio-integration/pull/109 requires on this to work. More functional details are in that PR.

As it is a servlet filter it works both for hawtio JVM and others via proxy


tadayosi (approved):

tadayosi:
LGTM. Thanks!

PR #2671, opened 2021-02-03
Support for proxy servlet broke jolokia. Fix tested in both scenarios

skarsaune:
Managed to break support for local jolokia servlet in last change for proxy.
Now works for both.
@tadayosi Apologies for managing to miss this in #2669


tadayosi (approved):

tadayosi:
Thanks again!

PR #2682, opened 2021-06-07
springboot example is not working 

PradeepNain:
Cannot resolve plugin io.hawt:hawtio-maven-plugin:2.14-SNAPSHOT

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.hawt</groupId>
     <artifactId>project</artifactId>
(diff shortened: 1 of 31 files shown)


oscerd (changes requested):
The example need to be in sync with the main pom

oscerd (changes requested):
It needs to be on 2.14-snapshot. 2.13.4 has been released already. You need to build the full project with 2.14-Snapshot as version. Once built the example will work.

PradeepNain:
sure, i'll check this.

PR #2741, opened 2022-05-16
Fix NPE in FlightRecordingDownloadFacade

moonfruit:
When `super.getPathInfo()` be null, this Request Wrapper will throw an NPE.


tadayosi (approved):
Thanks!

PR #2748, opened 2022-06-16
chore(console-assembly): update dependencies

GChuf:
Updates node (v10 -> v14) and yarn versions (v1.19 -> v1.22) and removed installation of npm (when downloaded through maven build)
Updates other dev dependencies
Speeds up build times (from 75s to 50s on my machine)
Gulp build process needed some editing due to node + gulp update

Let me know if anything needs changing before/if this gets merged


tadayosi:
The changes look good, but the build&test CI doesn't pass. Can you see why it fails?

tadayosi (approved):
Looks good. Let's merge this and see what happens.

tadayosi:
Thank you!

GChuf:
@tadayosi it seems github build times aren't so different after this commit.
Maybe it would make sense to include node_modules folder in cache, like you do for maven (.m2/repository).
It would save ~20seconds and downloading node dependencies on each commit.

PR #2793, opened 2023-02-08
camel-springboot-dependency fix

nstefeko:
Camel dependecies fixed. 

Springboot-keycloak is not working, example of errors in the screenshot

image


examples/springboot-log4j2/pom.xml
```diff
@@ -91,14 +91,17 @@
     <dependency>
       <groupId>org.apache.camel.springboot</groupId>
       <artifactId>camel-spring-boot-starter</artifactId>
+      <version>${camel-version}</version>
```
tadayosi, line 94: why do we need the versions? since we define bom above.

tadayosi:
> Springboot-keycloak is not working, example of errors in the screenshot

@nstefeko Do you say it starts not working since the fix you've done?  Or Doesn't it work even with the current main branch?

tadayosi:
@nstefeko Also, I don't like attaching a screenshot for reporting an issue, which isn't of much help, not to say it's low resolution.

tadayosi:
And please refer to the issue you are going to fix with this pull req from the commit comment, or in the pull req description.

nstefeko:
@tadayosi Hello I wasn't able to build the examples without version in
```
<dependency>
      <groupId>org.apache.camel</groupId>
      <artifactId>camel-debug</artifactId>
 </dependency>
```
but 
```
 <dependency>
      <groupId>org.apache.camel.springboot</groupId>
      <artifactId>camel-debug-starter</artifactId>
    </dependency>
```
works. Is that a better solution? Thank you for helping me

tadayosi:
Yes, for Spring Boot apps it should be better to use starter deps instead of direct deps.

tadayosi (approved):
lgtm, thanks!

PR #2798, opened 2023-02-22
[E2E Tests] Camel Endpoints and Routes tests

jsolovjo:
Adding E2E tests covering Camel Endpoints and Routes


tadayosi (changes requested):
It seems there's something wrong with maven surefire plugin settings with E2E tests.

However, I think what's more problematic here is that the Test workflow also runs E2E tests, where we already have E2E Test workflow for them.

tadayosi (approved):

tadayosi:
thanks!

Issue #2815, opened 2023-04-07
E2E tests are flaky


tadayosi:
Sometimes it fails, like this:
https://github.com/hawtio/hawtio/actions/runs/4593296569/jobs/8111604905

tadayosi:
Can you have a look? @jsolovjo

jsolovjo:
@tadayosi yes, I can have a look
I noticed that first run of e2e tests on jdk17 is failing, but the rerun of the job improves it

But I agree it shouldn't be like this, I will check it

tadayosi:
@jsolovjo It occasionally fails not only on JDK17 but also on JDK11. Retrying multiple times should fix it but it shouldn't be that flaky.

jsolovjo:
@tadayosi ok, I will check it tomorrow, I didn't know that it got worse, sorry

jsolovjo:
Done, the tests seem stabilised

tadayosi:
Flakiness is not yet fixed.
https://github.com/hawtio/hawtio/actions/runs/4675419767

tadayosi:
should be fixed with #2823 finally.

PR #2818, opened 2023-04-11
[E2E Tests] Stabilize open menu tabs method

jsolovjo:
The E2E tests became [flaky](https://github.com/hawtio/hawtio/issues/2815) and most failures were related to one method. By default, the tests are running in a background (headless mode) and most likely there were situations (connectivity issues like delays) when Selenide thought that some menu tabs were hidden under More tab (drop-down menu) and it was waiting for it. It was causing test failures and a job rerun was required. 

With this method deletion the situation should be stabilised.


jsolovjo:
Looks like it didn't help and the issue is different

jsolovjo:
Still trying to identify the issue. Locally everything works just fine without any issue, but on github it randomly fails.

jsolovjo:
Hopefully, this will be enough to stabilise the e2e tests

tadayosi (approved):
thanks!

PR #2851, opened 2023-06-06
[E2E Tests] Camel tree id-s tests and log out stabilization, version updates.

jsolovjo:
In this PR:

- Test frameworks version updates
- Log out action fix
- Camel tree id-s stabilization
- Added one check as at some point Hawtio is slower than Selenide and sometimes the content is not loaded properly

The github checks will fail (usually, 3-4 failing tests scenarios per runtime) because of https://github.com/hawtio/hawtio-next/issues/318

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -15,13 +15,13 @@
   <packaging>pom</packaging>
 
   <properties>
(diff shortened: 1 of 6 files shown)


tadayosi (approved):
thx!

PR #2854, opened 2023-06-09
Run hawtio via jbang

Croway:
The jbang app can be installed and used as follows:

```
➜  hawtio git:(Issue-2785-hawtio-jbang) ✗ jbang app install hawtio@hawtio/hawtio
➜  hawtio git:(Issue-2785-hawtio-jbang) ✗ hawtio
...
Embedded Hawtio: You can use --help to show usage
```

naming convention `hawtio@hawtio/hawtio` is not great, but this is how jbang scans repository internally.

@tadayosi let me know if hawtio-app should be removed, in case the code inside App.java can be moved to HawtioJBang, or I can create two submodules hawtio-jbang/hawtio-main and hawtio-jbang/hawtio-core, where core contains App and dependencies, and hawtio-main is the actual hawtio-jbang.


tadayosi (approved):
Great, thanks!

tadayosi:
@Croway Yes, if we provide Hawtio JBang then we don't need hawtio-app anymore. Ultimately we'd like to eliminate all the dependencies to Jetty from Hawtio, which means we should remove hawtio-app and hawtio-embedded. For now, it should be enough to remove hawtio-app.

Croway:
I had to reintroduce the assembly plugin in order to load hawtio-default from the classloader, I have the following questions but I think it is almost done:

* right now hawtio-embedded contains similar functionalities as hawtio-app in order to work with jbang, is it expected? as far as I understand hawtio-default is the only `implementation` that can be used, do you expect this to change?
* should I move hawtio-jbang module under deploy folder?
* when using `jbang app install` different locations are queried, in particular, in order to install hawtio with `jbang app install hawtio@hawtio`, the following project should be created under hawtio org `jbang-catalog` with `jbang-catalog.json` and `HawtioJBang.java` files, this means 1 more repository to maintain, in particular after each release the version should be updated in `HawtioJBang.java`, let me know if this is worth, otherwise `jbang app install hawtio@hawtio/hawtio` can be used

tadayosi:
> * right now hawtio-embedded contains similar functionalities as hawtio-app in order to work with jbang, is it expected? as far as I understand hawtio-default is the only `implementation` that can be used, do you expect this to change?

Hmm, I don't understand your question exactly. `hawtio-embedded` is for embedded usages of hawtio, and `hawtio-app` is a launcher that makes use of `hawtio-embedded` internally, so there should be no overlap in functonalities. `hawtio-default`, `hawtio-base`, and `hawtio-war` are the flavours of hawtio war-style distribution. For V3, I don't see much value in keeping `hawtio-default` and `hawtio-base`, and I'd like to investigate if we can remove them and provide only `hawtio-war`.

> * should I move hawtio-jbang module under deploy folder?

Let's keep it top level for now. We can later reconsider the location.

> * when using `jbang app install` different locations are queried, in particular, in order to install hawtio with `jbang app install hawtio@hawtio`, the following project should be created under hawtio org `jbang-catalog` with `jbang-catalog.json` and `HawtioJBang.java` files, this means 1 more repository to maintain, in particular after each release the version should be updated in `HawtioJBang.java`, let me know if this is worth, otherwise `jbang app install hawtio@hawtio/hawtio` can be used

I think `jbang app install hawtio@hawtio/hawtio` is enough for now. Later if we'll become more serious about the JBang installation method, we can then polish it by maintaining an extra repo for it.

Croway:
`there should be no overlap in functonalities` exactly, I see hawtio-app as a launcher and opinionated about the war flavour (hawtio-default), but since I removed hawtio-app, I needed a way to opinionate the flavour, that's why I added it to hawtio-embedded, right now if you run `hawtio` the hawtio-default is used OOB, but of course this can be customized with `hawtio --war=...` or `hawtio --war-location=...`. If hawtio-embedded don't have to be opinionated about the flavour, I am afraid that a new module should be added.

tadayosi (approved):
LGTM. Can we change it from draft to ready for review/merge?

Croway:
> LGTM. Can we change it from draft to ready for review/merge?

done! thank you

Issue #2855, opened 2023-06-13
Login page doesn't redirect if user is already logged in on Springboot


mmuzikar:
When a user is already logged in and navigates to what would be the login page (`/actuator/hawtio` on SB `/hawtio` on Quarkus) then they should be redirected to the app. Now that works only on Quarkus, on SB user is presented with blank screen.

Working on Quarkus as expected: 
![Peek 2023-06-13 09-49](https://github.com/hawtio/hawtio/assets/46345469/53917ea9-9b61-4235-8ab0-af5da5d46e51)

Not working on SB:
![Peek 2023-06-13 09-48](https://github.com/hawtio/hawtio/assets/46345469/793e3df1-1ebf-41ee-a8ae-4a3802a6f551)

PR #2860, opened 2023-06-16
Add redirect to hawtio/

Croway:
This PR fixes the issue https://github.com/hawtio/hawtio/issues/2855 and applies same behavior as quarkus app (redirect in case of missing / in case of `actuator/hawtio`). Though I am not sure if there is another (better) way to achieve the same result.


Croway:
Hi @tadayosi thanks for the review, I did not know about the context root configuration, I hope it works as expected with the latest changes.

tadayosi (approved):
Wow, that's simple. Thanks for your great contribution!

PR #2872, opened 2023-06-28
Jetty 10 Upgrade for 2.x

Croway:
In addition to the cherry-pick from 3.x branch, I had to add the following:

- Explicitly set `maven-plugin-plugin` version because it was clashing with jetty 10
- Add `javax.servlet-api` to hawtio-embedded, given that is not provided by jetty 10 (I wonder why though)


tadayosi (approved):
thanks, awesome!

PR #2873, opened 2023-06-30
Use esbuild plugin to speedup JS builds

mmuzikar:
From testing on my machine this change speeds up prod builds from around 2 minutes to around 30 seconds.

I noticed @tadayosi encountered the same issue as me https://github.com/module-federation/universe/issues/827, so I used the suggested workaround


tadayosi (approved):
great, thanks!

tadayosi:
@mmuzikar could you apply the same fix to hawtio-next, too?

PR #2877, opened 2023-07-07
gh-2875 fix workflow reporting on push events

### LINKED ISSUE #2875: E2E tests failing at publish-results job
Unfortunately, merging the pull req didn't fix the E2E failures:
https://github.com/hawtio/hawtio/actions/runs/5452977928

We need to find and fix the root cause of why it happens. @mmuzikar Could you have a look?

mmuzikar:
Closes #2875


tadayosi (approved):

tadayosi:
@mmuzikar Looks like it stil fails at the last stage.

mmuzikar:
Yes - I'm looking at it, however it doesn't fail on my fork so I have to debug it here :/

tadayosi:
So, it could be related to the permissions on the hawtio project.

.github/workflows/e2e_test.yml
```diff
@@ -112,9 +112,7 @@ jobs:
           key: e2e-docker-images-${{ matrix.java }}-${{ github.run_id }}-${{ github.run_attempt }}
 
   publish-results:
-    permissions:
-      checks: write
-      pull-requests: write
+    permissions: write-all
```
tadayosi, line 115: Should we still need this?  Can't we revert to the original?  The less permissive, the better.

tadayosi:
Could be this?
https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

tadayosi:
So, how should we proceed with this?

mmuzikar:
This change should now have access to the secrets and checkout the correct branch... but due to how the `pull_request_target` works the tests won't run in this PR

mmuzikar:
@tadayosi here you can see the workflow working: https://github.com/hawtio/hawtio/pull/2889 and push works as well: https://github.com/hawtio/hawtio/actions/runs/5575141223

tadayosi:
OK, let's merge then.

PR #2900, opened 2023-07-20
[E2E Tests] New Camel Contexts Operations and Chart test scenarios

jsolovjo:
### In this PR:

New test scenarios for Camel Context covering:
- Camel Chart
  - View Chart
  - Edit Chart

- Camel Operatios
  - Execute the operation returning a result of `boolean` type
  - Execute the operation returning a result of `int` type

### DIFF
--- a/tests/hawtio-test-suite/src/main/java/io/hawt/tests/features/pageobjects/fragments/camel/tabs/common/CamelChart.java
+++ b/tests/hawtio-test-suite/src/main/java/io/hawt/tests/features/pageobjects/fragments/camel/tabs/common/CamelChart.java
@@ -1,0 +1,121 @@
+package io.hawt.tests.features.pageobjects.fragments.camel.tabs.common;
+
+import static com.codeborne.selenide.CollectionCondition.sizeGreaterThan;
+import static com.codeborne.selenide.Condition.enabled;
+import static com.codeborne.selenide.Condition.exist;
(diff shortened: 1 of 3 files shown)


tadayosi (approved):
LGTM, but the tests are failing. Please check them.

jsolovjo:
> Details

Thanks. I will check it. Locally, it worked fine - I will investigate what went wrong.

tadayosi:
We released 3.0 Beta1 already, so please rebase it first.

jsolovjo:
Hi @tadayosi 
I will rebase it and I am thinking about to remove one test scenario as it it quite flaky - to get the value of the chart you have to hover over the specific chart bar. The thing is it's even buggy when checking it manually - quite often it doesn't show up from the first hover and it refreshes every a few seconds, to it loses the focus. I will try a few more attempts and if no success, I will leave a test checking that there are actually chart bars.

tadayosi:
> The thing is it's even buggy when checking it manually - quite often it doesn't show up from the first hover and it refreshes every few seconds, so it loses the focus. I will try a few more attempts and if no success, I will leave a test checking that there are actually chart bars.

Yes, as the nature of the Chart view it refreshes frequently. The test should use a different approach than other views and might need to relax some checking. If possible, we can just check that the view constantly refreshes with different data.

tadayosi (approved):
Great job!  Thank you!

PR #2902, opened 2023-07-21
ci: Publish e2e docker images into quay.io

mmuzikar:
Builds test images on push and pushes them to Quay to be reused in hawtio-next.

Add tags to skip tests that require configuration that is not done in hawtio-next tests.
Related to https://github.com/hawtio/hawtio-next/pull/457

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -14,6 +14,9 @@ env:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        java: ['11', '17']
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout code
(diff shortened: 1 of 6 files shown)


.github/workflows/image.yml
```diff
@@ -43,4 +47,10 @@ jobs:
           PASSWORD: ${{ secrets.QUAY_PASSWORD }}
         run: |
           docker login -u $USERNAME -p $PASSWORD quay.io
-          docker push quay.io/hawtio/hawtio-example-quarkus:${{ env.project_version }}
```
tadayosi, line 46: you removed the example image, but we still use it.

tadayosi (changes requested):

.github/workflows/image.yml
mmuzikar: it gets pushed in the for loop below with the rest of the images

tadayosi: ah, ok. then we dont need the previous step of setting the project version to env, correct? please remove it then.

mmuzikar: removed the step

PR #2904, opened 2023-07-21
hawtio-jbang - print version

Croway:
Remove warning from slf4j, and print version with `hawtio --version`.


tadayosi (approved):
Awesome, thanks!

PR #2905, opened 2023-07-21
Bump javax servlet and karaf version

Croway:
jetty 10 need javax.servlet 4.0.1, and the karaf version we had was too old for javax.servlet 4.0.1.


oscerd (approved):

tadayosi (approved):
👍

PR #2908, opened 2023-07-28
Upgrade to jetty12.0.0beta3 and spring-boot 3 

mmelko:
I've upgraded the version of sb to sb3. I'm still facing the problems with resolving index.html and other things.

1.  It looks like links in static files aren't properly regenerated (probably responsibility of  `ResourceUrlEncodingFilter`) when management.server.port is configured to the different port as the server.port. When management.server.port isn't configured hawtio application is properly resolved.

2. application is resolved only when  directly  index.html is requested. When `actuator/hawtio/` is requested then I'm getting 404. This probably can be fixed by manual redirect to index.html in HawtioEndpoint but I don't feel that's the best solution. Probably the ResourceHandlers has to be tweaked.

### DIFF
--- a/deploy/hawtio-embedded/src/main/java/io/hawt/embedded/Main.java
+++ b/deploy/hawtio-embedded/src/main/java/io/hawt/embedded/Main.java
@@ -17,6 +17,14 @@
  */
 package io.hawt.embedded;
 
+import org.eclipse.jetty.ee10.webapp.WebAppContext;
+import org.eclipse.jetty.server.*;
+import org.eclipse.jetty.server.handler.ContextHandlerCollection;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
+
 import java.awt.*;
 import java.io.File;
 import java.io.IOException;
@@ -25,20 +33,6 @@ import java.net.InetSocketAddress;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.concurrent.Callable;
-
-import org.eclipse.jetty.server.Connector;
-import org.eclipse.jetty.server.HttpConfiguration;
-import org.eclipse.jetty.server.HttpConnectionFactory;
-import org.eclipse.jetty.server.Server;
-import org.eclipse.jetty.server.ServerConnector;
-import org.eclipse.jetty.server.SslConnectionFactory;
-import org.eclipse.jetty.server.handler.HandlerCollection;
-
-import org.eclipse.jetty.util.ssl.SslContextFactory;
-import org.eclipse.jetty.webapp.WebAppContext;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import picocli.CommandLine;
 
 /**
  * A simple way to run hawtio embedded inside a JVM by booting up a Jetty server
@@ -138,7 +132,7 @@ public class Main implements Callable<Integer>  {
     public String run(boolean join) throws Exception {
         Server server = new Server(new InetSocketAddress(InetAddress.getByName(host), port));
 
-        HandlerCollection handlers = new HandlerCollection();
+        ContextHandlerCollection handlers = new ContextHandlerCollection();
         handlers.setServer(server);
         server.setHandler(handlers);
 
@@ -212,7 +206,9 @@ public class Main implements Callable<Integer>  {
         webapp.setParentLoaderPriority(true);
         webapp.setLogUrlOnStart(true);
         webapp.setInitParameter("scheme", scheme);
-        webapp.setExtraClasspath(extraClassPath);
+        if (extraClassPath!= null) {
+            webapp.setExtraClasspath(extraClassPath);
+        }
         return webapp;
     }
 
@@ -246,7 +242,7 @@ public class Main implements Callable<Integer>  {
         return scheme;
     }
 
-    protected void findThirdPartyPlugins(Logger log, HandlerCollection handlers, File tempDir) {
+    protected void findThirdPartyPlugins(Logger log, ContextHandlerCollection handlers, File tempDir) {
         File dir = new File(plugins);
         if (!dir.exists() || !dir.isDirectory()) {
             return;
@@ -263,7 +259,7 @@ public class Main implements Callable<Integer>  {
             war -> deployPlugin(war, log, handlers, tempDir));
     }
 
-    private void deployPlugin(File war, Logger log, HandlerCollection handlers, File tempDir) {
+    private void deployPlugin(File war, Logger log, ContextHandlerCollection handlers, File tempDir) {
         String contextPath = resolveContextPath(war);
 
         WebAppContext plugin = new WebAppContext();
--- a/deploy/hawtio-embedded/src/main/java/io/hawt/embedded/Main.java
+++ b/deploy/hawtio-embedded/src/main/java/io/hawt/embedded/Main.java
@@ -17,14 +17,6 @@
  */
 package io.hawt.embedded;
 
-import org.eclipse.jetty.ee10.webapp.WebAppContext;
-import org.eclipse.jetty.server.*;
-import org.eclipse.jetty.server.handler.ContextHandlerCollection;
-import org.eclipse.jetty.util.ssl.SslContextFactory;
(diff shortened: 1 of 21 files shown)


deploy/hawtio-embedded/src/main/java/io/hawt/embedded/Main.java
```diff
@@ -17,6 +17,14 @@
...
 package io.hawt.embedded;
 
+import org.eclipse.jetty.ee10.webapp.WebAppContext;
+import org.eclipse.jetty.server.*;
+import org.eclipse.jetty.server.handler.ContextHandlerCollection;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
+
```
tadayosi, line 27: Nitpicky but we prefer to place `java.*` and `javax.*` earlier than the rest. If you haven't changed the preferences of imports in your IDE, please change it for the hawtio project and fix the ordering.

tadayosi: For example, in IDEA it should look like this:

image

In doing so, we can put `jakarta.*` between `javax.*` and all the other imports.

hawtio-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd"
-         version="3.0">
+         xsi:schemaLocation = "https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_5_0.xsd"
+         version="5.0">
```
tadayosi, line 5: Why is this still 5.0?

platforms/springboot/src/integration-test/java/io/hawt/springboot/HawtioSpringBootTestSupport.java
```diff
@@ -244,8 +244,8 @@ public TestPropertiesBuilder servletPath(String servletPath) {
             return this;
         }
 
-        public TestPropertiesBuilder managementBasePath(String managementBasePath) {
-            this.managementBasePath = managementBasePath;
+        public TestPropertiesBuilder managemenBasePath(String managementContextPath) {
```
tadayosi, line 247: Again please fix the typo: `managemen` -> `management` ;-)

platforms/springboot/src/main/java/io/hawt/springboot/HawtioManagementConfiguration.java
```diff
@@ -34,23 +35,21 @@
...
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.core.Ordered;
 import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
 import org.springframework.web.servlet.mvc.AbstractUrlViewController;
 
 import static io.hawt.web.filters.BaseTagHrefFilter.PARAM_APPLICATION_CONTEXT_PATH;
 
-@ManagementContextConfiguration
+@Configuration
```
tadayosi, line 52: Isn't this the reason why we have the issue with the management feature?  If `@ManagementContextConfiguration` isn't simply available at SB 3, maybe we can track the changes of the deletion of the annotation to see why it's removed and how we should rewrite it for our app.

platforms/springboot/src/main/java/io/hawt/springboot/JolokiaProperties.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "management.endpoint.jolokia")
+public class JolokiaProperties {
+
+    /**
+     *
```
tadayosi, line 12: any javadoc comments? or simply remove the comment block.

platforms/springboot/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
```diff
@@ -0,0 +1,4 @@
+io.hawt.springboot.JolokiaEndpointAutoConfiguration
+io.hawt.springboot.HawtioEndpointAutoConfiguration
+io.hawt.springboot.HawtioLogAutoConfiguration
+io.hawt.springboot.HawtioManagementConfiguration
```
tadayosi, line 4: I'm not knowedgeable about SB 3, but don't we also need the `imports` file for the management feature?

tadayosi (changes requested):
Overall, it looks great!

platforms/springboot/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
tadayosi: See for example:
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.actuate.autoconfigure.web.ManagementContextConfiguration.imports

hawtio-war/src/main/webapp/WEB-INF/web.xml
mmelko: I had to missed this when I tried jetty 11. Thanks!

platforms/springboot/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
mmelko: this!! Thanks. :)

platforms/springboot/src/main/java/io/hawt/springboot/HawtioManagementConfiguration.java
mmelko: Looks like we had a problem because I was missing imports for management as you mentioned. https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/actuate/autoconfigure/web/ManagementContextConfiguration.html I've tried with also with `@Configuration` only and suddenly it works

mmelko:
I updated the PR. Currently there is only problem with reaching directly .../actuator/hawtio/ -> it ends with server 500. Otherwise it looks like it's working. Also tests are passing locally with locally built jolokia with jakarta

hawtio-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd"
-         version="3.0">
+         xsi:schemaLocation = "https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_5_0.xsd"
```
tadayosi, line 4: I think you wanted to change to `web-app_6_0` instead of `web-app_5_0`

tadayosi (approved):
Only one minor change request, but otherwise it looks great. Thanks!

tadayosi:
> I updated the PR. Currently there is only problem with reaching directly .../actuator/hawtio/ -> it ends with server 500.

If it's not easy to fix, you can file another issue and address it separately from this base pull req.

Also, it's about time to move it from WIP to Ready for Merge.

Issue #2819, opened 2023-04-17
Support Spring Boot 3, Quarkus 3, Jakarta EE 10, and Camel 4


tadayosi:
Since SB 3 and Quarkus 3 only support Java 17 and Jakarta EE 10, and Camel 4 will be based on SB 3 and Quarkus 3, for Hawtio to support all of them it'll need to upgrade all the stack as well. It would mean that it's probably not possible to support SB 2 and 3, or Camel 3 and 4 at the same time in a single version of Hawtio, unless we could find a nice way to support multiple verions of JEE and SB, Quarkus in a single project.

Currently we are developing Hawtio v3, but we might need to start Hawtio v4 development as well, which will be based on Java 17, Jakarta EE 10, SB 3, and Quarkus 3.

Relates #2817

Tasks are broken down into following:
- [x] https://github.com/hawtio/hawtio/issues/2866
- [x] https://github.com/hawtio/hawtio/issues/2867
- [x] https://github.com/hawtio/hawtio/issues/2868
- [x] https://github.com/hawtio/hawtio/issues/2869

tadayosi:
Done.

PR #2916, opened 2023-08-10
refactor(examples): spring-security example to work with spring-boot 3 and spring-security 6

mmelko:
This PR reflects changes of spring-security 6. Mainly moving out of `WebSecurityConfigurerAdapter` and change in csrf token behaviour that mitigates BREACH vulnerability and by default might not work with JS applications
related to #2819


tadayosi (changes requested):
In Java, it's not common to define multiple top-level classes in a single file. Rather, they should be defined as inner static classes inside the original one.

mmelko:
> In Java, it's not common to define multiple top-level classes in a single file. Rather, they should be defined as inner static classes inside the original one.

I updated the configuration file. Hope it's fine now :)

tadayosi (approved):

PR #2926, opened 2023-08-25
fix(e2e): fix branding styles loading with Spring Boot test application

tadayosi:


tadayosi:
cc @mmuzikar

tests/springboot/src/main/resources/hawtio-static/hawtconfig.json
```diff
@@ -3,8 +3,8 @@
     "appName": "Hawtio Spring Boot 2 Sample app for E2E testing purposes",
     "appLogoUrl": "img/hawtio-logo.svg",
     "companyLogoUrl": "img/hawtio-logo.svg",
```
mmuzikar, line 5: What about these logos @tadayosi?

tadayosi: You can copy them from the console itself but unless you provide your own they just point to the builtin img resources.

mmuzikar (approved):

PR #2928, opened 2023-08-29
fix(plugin): fix plugin needing hardcoded plugin url config

### LINKED ISSUE #348: Allow plugins to be served without hardcoding URLs
Currently user is required to specify URL for the plugin config, so the URLs are hardcoded to use one given port. This is an issue when running the app in a docker container with a mapped port. Then the assigned port can differ from the port that's specified in the app config. 
When running hawtio in a docker container with mapped ports everything works fine except for the plugins. So it should be possible to omit the URL for it to be served from the same endpoint as the hawtio app is running.

kariuwu:
closes [#348](https://github.com/hawtio/hawtio-next/issues/348)

### DIFF
--- a/platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
+++ b/platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
@@ -26,10 +26,30 @@ public class HawtioPlugin {
     private Boolean bustRemoteEntryCache;
     private String pluginEntry;
 
-    public HawtioPlugin(String url, String scope, String module) {
-        this.url = url;
+    /**
+     * Use a plugin from remote repository.
+     * These are the parameters required to load a remote Hawtio plugin (a.k.a. Module Federation remote module):
+     *
+     * @param url The URL of the remote entry for the plugin. This must be the same location as the Hawtio console.
+     * @param scope The name of the container defined at Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     * @param module The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     */
+    public HawtioPlugin(String scope, String module, String url) {
         this.scope = scope;
         this.module = module;
+        this.url = url;
+    }
+
+    /**
+     * Use a plugin that's located in the application itself.
+     * This constructor automatically uses address of the application.
+     * These are the parameters required to load a remote Hawtio plugin (a.k.a. Module Federation remote module):
+     *
+     * @param scope The name of the container defined at Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     * @param module The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     */
+    public HawtioPlugin(String scope, String module) {
+        this(scope, module, "");
     }
 
     @JsonProperty("url")
--- a/tests/quarkus/src/main/resources/application.properties
+++ b/tests/quarkus/src/main/resources/application.properties
@@ -27,7 +27,8 @@ quarkus.security.users.embedded.users.hawtio = hawtio
 quarkus.security.users.embedded.roles.hawtio = admin
 
 # 'sample' custom plugin settings
-quarkus.hawtio.plugin.sample-plugin.url = http://localhost:8081
+#Uses the application URL to serve the plugin
+#quarkus.hawtio.plugin.sample-plugin.url = http://localhost:8080
 quarkus.hawtio.plugin.sample-plugin.scope = samplePlugin
 quarkus.hawtio.plugin.sample-plugin.module = ./plugin
 
--- a/tests/springboot/src/main/java/io/hawt/tests/spring/boot/SpringBootService.java
+++ b/tests/springboot/src/main/java/io/hawt/tests/spring/boot/SpringBootService.java
@@ -36,12 +36,10 @@ public class SpringBootService {
         /*
          * These are the parameters required to load a remote Hawtio plugin (a.k.a. Module Federation remote module):
          *
-         * - url: The URL of the remote entry for the plugin. This must be the same location as the Hawtio console.
          * - scope: The name of the container defined at Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
          * - module: The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
          */
         HawtioPlugin plugin = new HawtioPlugin(
-            "",
             "samplePlugin",
             "./plugin");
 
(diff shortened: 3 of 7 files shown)


platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -16,6 +16,7 @@
  * <a href="https://github.com/module-federation/universe/blob/utils-1.4.0/packages/utilities/src/utils/importRemote.ts#L9-L15">ImportRemoteOptions</a>
  * interface defined in @module-federation/utilities package.
  */
+
```
tadayosi, line 19: Why this new line?

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -26,12 +27,31 @@ public class HawtioPlugin {
     private Boolean bustRemoteEntryCache;
     private String pluginEntry;
 
+    /**
+     *
```
tadayosi, line 31: If you want to add javadoc, please also add the comment body from the first line.

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -26,12 +27,31 @@ public class HawtioPlugin {
...
+     * @param module The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     */
     public HawtioPlugin(String url, String scope, String module) {
         this.url = url;
         this.scope = scope;
         this.module = module;
     }
 
+    /**
+     *
```
tadayosi, line 43: same here

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -26,12 +27,31 @@ public class HawtioPlugin {
...
+     * @param scope The name of the container defined at Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     * @param module The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     *
+     * This constructor automatically uses hawtio address.
```
tadayosi, line 47: It's not commont to write the comment body at the bottom. Please move it to the top.

Also, it's not precisely "hawtio address" but the address of the application.

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -26,12 +27,31 @@ public class HawtioPlugin {
...
+     */
+    public HawtioPlugin(String scope, String module){
```
tadayosi, line 49: insert space before the curly brace.

tadayosi, line 49: I don't think it's a good API design that the two constructors have different meanings for the arguments with the same positions and type.
```
HawtioPlugin(String url, String scope, String module)
HawtioPlugin(String scope, String module)
```
If we want to provide such two constructors, I think it's better to move the position of `url` to the end for the first one. When doing so, make sure to update all the places where the original constructor is used.

Also, the 2nd one should be implemented using the 1st (most versatile) one like this:
```
    this(scope, module, "");
```

tadayosi (changes requested):
Thanks!  There are a few comments to be addressed. Please have a look.

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -26,10 +26,28 @@ public class HawtioPlugin {
     private Boolean bustRemoteEntryCache;
     private String pluginEntry;
 
-    public HawtioPlugin(String url, String scope, String module) {
-        this.url = url;
+    /**
+     * These are the parameters required to load a remote Hawtio plugin (a.k.a. Module Federation remote module):
```
tadayosi, line 30: This is ok for the second line of the constructor javadoc comment, but the first line should first explain what does this constructor is meant to do. Explain the constructor itself rather than the parameters it accepts. The parameters are self-evident in the `@param` comments already.

See example:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ArrayList.html#%3Cinit%3E()

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
tadayosi: Again, explain what this constructor does.

tests/quarkus/src/main/resources/application.properties
```diff
@@ -27,7 +27,7 @@ quarkus.security.users.embedded.users.hawtio = hawtio
 quarkus.security.users.embedded.roles.hawtio = admin
 
 # 'sample' custom plugin settings
-quarkus.hawtio.plugin.sample-plugin.url = http://localhost:8081
+#quarkus.hawtio.plugin.sample-plugin.url = http://localhost:8080
```
tadayosi, line 30: Maybe it's better to add a comment above the commented line about when this parameter should be used.

platforms/springboot/src/main/java/io/hawt/springboot/HawtioPlugin.java
```diff
@@ -26,10 +26,28 @@ public class HawtioPlugin {
...
+
+    /**
+     * This constructor automatically uses address of the application.
+     * These are the parameters required to load a remote Hawtio plugin (a.k.a. Module Federation remote module):
+     *
+     * @param scope The name of the container defined at Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     * @param module The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
+     */
+    public HawtioPlugin(String scope, String module) {
+        this(scope,module, "");
```
tadayosi, line 50: add space between `scope` and `module`

tests/springboot/src/main/java/io/hawt/tests/spring/boot/SpringBootService.java
```diff
@@ -41,7 +41,6 @@ public HawtioPlugin samplePlugin() {
          * - module: The path exposed from Webpack ModuleFederationPlugin. See also: sample-plugin/craco.config.js
          */
         HawtioPlugin plugin = new HawtioPlugin(
-            "",
```
tadayosi, line 44: Now you omitted this parameter, so should update the comment above accordingly (i.e. `url`).

tadayosi (changes requested):
A bit more request for changes.

tadayosi (approved):

tadayosi:
there's one more place to update:
https://github.com/hawtio/hawtio/blob/3.x/examples/springboot-plugin/src/main/java/io/hawt/example/spring/boot/SampleSpringBootService.java#L27-L30
but I'll take care of it.

PR #2937, opened 2023-09-11
[E2E Tests] @notHawtioNext tags in About window fixed

nstefeko:
- Some tests were skipped because there were issues between springboot and quarkus, or there were just issues during the refactor. I fixed the tests with @notHawtioNext tags for About modal window feature


tadayosi (approved):
thanks!

PR #2939, opened 2023-09-14
[E2E Tests] hawng-87 fix the tests

jsolovjo:
Hello,

this PR should fix test failures on https://github.com/hawtio/hawtio-next/pull/554

FYI @nstefeko I had to modify the tests and hawtio config files. Locally, it works fine, but on PRs the tests fail due to different texts in About dialog window. So this PR should fix it.

As an example, I checked the About dialog which shows up during E2E tests on PRs and it differs from what we test.


tadayosi (approved):
thanks!

nstefeko:
Thank you for explaining @jsolovjo

PR #2955, opened 2023-10-11
Elytron authentication support

rnetuka:
Elytron support for Hawtio authentication

Wildfly has two options - JAAS or Elytron. On JDK 17, there is however only the Elytron option. These changes introduce a new ElytronAuthenticator packed in the platforms/wildfly module. It is read dynamically by Java's ServiceLoader.

The old Authenticator was renamed into JaasAuthenticator (since I needed its name for a new interface) and an abstract class was created for these two to share some common methods. All authentication should be now called thru AuthenticationManager, which will select the correct authenticator for the work.

Tested on both Wildfly with legacy authentication (uses JaasAuthenticator), and with Elytron (uses ElytronAuthenticator). On other servers, the ElytronAuthenticator shouldn't be there, since it is only packed with hawtio-wildfly.war.


tadayosi:
For some reason, the E2E tests don't run successfully.

@jsolovjo It looks like E2E doesn't work with the `main` branch. Could you have a look at it?  We'd like to verify this pull req on the `main` branch with E2E because it's an important change to apply.

jsolovjo:
> For some reason, the E2E tests don't run successfully.
> 
> @jsolovjo It looks like E2E doesn't work with the `main` branch. Could you have a look at it? We'd like to verify this pull req on the `main` branch with E2E because it's an important change to apply.

Hi @tadayosi 
Sure, I will have a look at this

tadayosi:
@jsolovjo Thanks!  Ah, is it the exact issue of Selenium we discussed some time ago?

jsolovjo:
> @jsolovjo Thanks! Ah, is it the exact issue of Selenium we discussed some time ago?

Most likely, yes
I will try to fix it now

tadayosi:
@rnetuka CI is now fixed. Could you rebase it and rerun the check please?

tadayosi (approved):
Thanks!

chencao187:
Huge thanks for this fix.

I think I found a issue with this change. 
When using JDK 8 there would be a issue regarding the Wildfly version:

15:30:22,620 ERROR [io.undertow.request] (default task-1) UT005023: Exception handling request to /hawtio/auth/login: java.lang.UnsupportedClassVersionError: Failed to link org/jboss/as/webservices/util/SubjectUtil (Module "deployment.hawtio-wildfly-2.17.7.war" from Service Module Loader): org/jboss/as/webservices/util/SubjectUtil has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method) [rt.jar:1.8.0_202]
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763) [rt.jar:1.8.0_202]
	at java.lang.ClassLoader.defineClass(ClassLoader.java:839) [rt.jar:1.8.0_202]
	at org.jboss.modules.ModuleClassLoader.doDefineOrLoadClass(ModuleClassLoader.java:351) [:2.0.2.Final]
	at org.jboss.modules.ModuleClassLoader.defineClass(ModuleClassLoader.java:482) [:2.0.2.Final]
	at org.jboss.modules.ModuleClassLoader.loadClassLocal(ModuleClassLoader.java:276) [:2.0.2.Final]
	at org.jboss.modules.ModuleClassLoader$1.loadClassLocal(ModuleClassLoader.java:79) [:2.0.2.Final]
.................................

I think the problem is that in the pom.xml file, the Wildfly version is increased to 29.0.1.

`    <wildfly-integration-version>29.0.1.Final</wildfly-integration-version>`

----------------------------------------------------------------------------------------------------
The last version of Wildfly that support JDK 8 is 26.1.3. So this Hawtio 2.17.7 does not run properly on Wildfly 26.1.3 or below.

I tried change the wildfly-tetegration-version to 26.1.3 and build it manually for Wildfly. Problem solved.
I check the JDK compatibility of the Hawtio: https://hawt.io/docs/versions.html
Sadly it's not specified for Hawtio 2.x.x
So I hope there could be at least some clarification on the JDK compatibility.

tadayosi:
Hawtio 2.x had been Java >=11 only for long time.

PR #3044, opened 2023-11-16
build: Bump Node versions update for maven frontend plugin

mmuzikar:
This is needed as with older Node versions connection through IPv6 was unstable. This should fix the issue. 
Btw @tadayosi what do you think about moving these properties to the parent pom so we can upgrade the versions in just one place?

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -13,11 +13,6 @@
   <description>Hawtio :: ${project.artifactId}</description>
   <packaging>pom</packaging>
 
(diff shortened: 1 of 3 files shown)


oscerd (approved):

tadayosi (approved):
Sounds good to me. Could you pull them up to top pom.xml with this?

tadayosi:
and once it's merged please backport it to 3.x as well.

PR #3069, opened 2023-11-23
chore(springboot-example): Add jmx exposure comment for the spring-boot plugin

mmelko:


tadayosi (approved):

tadayosi:
@mmelko Thanks. Is it only relevant to v4?  I guess it's not needed for Spring Boot 2.x, correct?

mmelko:
@tadayosi yes. It default exposures are different in spring-boot 2.x However if you want to expose `Traces` you need to expose `Httptrace` too.  We didn't have exposed it before.

PR #3102, opened 2023-12-04
[E2E Tests] Fix add endpoint from Data test

jsolovjo:
In this PR:
- Fix Add an endpoint from Data test scenario which started failing in https://github.com/hawtio/hawtio-next/pull/695


tadayosi (approved):
thx!

PR #3181, opened 2024-01-16
fix(quarkus): sanitize path in LoadFromHawtioStatic

mmelko:


tadayosi (approved):

tadayosi:
@mmelko Tests are failing. It seems to be related to how Quarkus extension is built with classpaths.

PR #3184, opened 2024-01-19
feat(test): Add Hawtio Online tests & e2e tests on Hawtio Online

mmuzikar:
Adds possibility to run E2E testsuite with Hawtio Online. By default these tests are disabled and are enabled only by specifying ` -Dio.hawt.test.use.openshift=true -Dio.hawt.test.openshift.index.image=...`. 
The Online tests are executed only when invoked to run on Openshift along with some specific tests for Hawtio Online. 
I refactored away the notion of smoke and all tests as we always invoked the all tests and tags are now used only for skipping tests that don't match the current environment. 
Currently covered areas for Hawtio Online are: testing cluster/namespace discovery of pods, smoke testing the discover page (all links link to correct OCP resources, changing the state of pods is reflected in the view, scaling).


mmuzikar:
One of the tested functionalities is running the E2E testsuite on Hawtio Online based instance. The Hawtio Online tests require a lot of the test infrastructure that's already set up in this codebase and splitting it in two would only cause more friction than benefit IMO.

tadayosi:
@mmuzikar I can understand it's **efficient** to have it here but still I'm not convinced that it's logically the correct place. You said we want to run the same suite for hawtio-online, but I think we've been doing that at hawtio-next as well. I don't see why we can't just duplicate the same settings at hawtio-online.

If you're concerned about the release schedule, then we could just run it privately elsewhere this time for hawtio-online. IMO, at least we should've exhausted discussion before choosing a bit eccentric place for the e2e.

tadayosi:
Also, since we are using `pull_request_target:` for the e2e workflow we couldn't verify if this big change really runs fine here until merging the big pull req.

tests/hawtio-test-suite/src/main/java/io/hawt/tests/features/config/TestConfiguration.java
```diff
@@ -47,23 +66,30 @@ public static boolean browserHeadless() {
     }
 
     public static AppDeployment getAppDeploymentMethod() {
+        if (deployment != null) {
+            return deployment;
+        }
         if (hasProperty(APP_URL)) {
-            return new URLDeployment(getRequiredProperty(APP_URL));
+            return deployment = new URLDeployment(getRequiredProperty(APP_URL));
```
tadayosi, line 73: This notation looks a bit strange for Java. Maybe you can just split lines if you really want to assign an object to `deployment` before return. Same for the following lines.

tadayosi:
Another concern with having it here is that then the e2e workflow becomes heavier with unnecessary tests that don't add value to the codebase by checking them here. That would drag other daily contributions from running the tests otherwise more quickly.

And one question with the pull req. How do you specify the version of hawtio-online?  Ah, is this pull req only to add scenarios for hawtio-online that are intended to be run at that repository later just like hawtio-next?  If so, what's really lacking would be clearer and more detailed explanation on what you intend to add here.

tests/pom.xml
```diff
@@ -24,6 +24,7 @@
     <cucumber-version>7.14.0</cucumber-version>
     <testcontainers-version>1.19.3</testcontainers-version>
     <awaitility-version>4.2.0</awaitility-version>
+    <hawtio.bom.version>${project.version}</hawtio.bom.version>
```
tadayosi, line 27: why do we need this? please explain.

tests/quarkus/pom.xml
```diff
@@ -63,80 +57,60 @@
...
     <dependency>
       <groupId>org.apache.camel.quarkus</groupId>
       <artifactId>camel-quarkus-debug</artifactId>
     </dependency>
-
-    <!-- Quarkus -->
+    <dependency>
+      <groupId>org.graalvm.sdk</groupId>
+      <artifactId>graal-sdk</artifactId>
+    </dependency>
```
tadayosi, line 75: do we really need this? I remember it's no longer necessary but is it yet in product?

tests/quarkus/pom.xml
```diff
@@ -38,15 +34,13 @@
...
-        <version>${project.version}</version>
+        <version>${hawtio.bom.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
     </dependencies>
   </dependencyManagement>
-
   <dependencies>
-    <!-- Camel Quarkus -->
```
tadayosi, line 49: Is there a reason you remove these comments?  I guess you ran the quarkus command which automatically removes them. Maybe we might want to restore them.

tests/springboot/pom.xml
```diff
@@ -160,6 +166,12 @@
       <artifactId>hawtio-example-sample-plugin</artifactId>
       <version>${project.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>fabric8-maven-plugin</artifactId>
+      <version>4.4.1</version>
```
tadayosi, line 173: maybe we should parameterise it somewhere, preferably the root `pom.xml`

tadayosi:
> I'm not sure I understand, the test code was executed in the actions and there's no change to the workflows which would affect the actions.
> 
> As for the placement it's just my preference to have all the code base in one place as much of the code is reused in different places and I wasn't sure if the time investment into splitting the codebase is worth it, as we will soon start to cross into hawtio-operator territory. The Hawtio Online parts are ignored here in the e2e, and this codebase would be pulled just like in hawtio-next for operator and online. I have the downstream automation setup to run this testsuite on OCP so if we end up deciding that each project should get its own testsuite I can do it this release.

Yeah, that's because you haven't given a clear explanation on what you're going to do with it from the beginning. So there are miscommunication and misunderstanding. It's still not too later and good if you can give us your original plan with it in more details so reviewers could have a better understanding on what's going to happen here.

tadayosi (approved):
Looks great now. Thanks!

PR #3186, opened 2024-01-23
feat(jbang): support red hat maven repositories for jbang

tadayosi:
This should enable users to run a Hawtio version provided on MRRC as follows:

```console
jbang -Dhawtio.jbang.version=4.0.0.redhat-00029 hawtio@hawtio/hawtio
```

@Croway Please review this to check if I'm doing it correctly.

### DIFF
--- a/hawtio-jbang/dist/HawtioJBang.java
+++ b/hawtio-jbang/dist/HawtioJBang.java
@@ -19,7 +19,7 @@
  */
 
 //JAVA 11+
-//REPOS mavencentral
+//REPOS mavencentral,redhat-ga=https://maven.repository.redhat.com/ga,redhat-ea=https://maven.repository.redhat.com/earlyaccess/all
 //DEPS io.hawt:hawtio-embedded:${hawtio.jbang.version:4.0-beta-1}
 
 package main;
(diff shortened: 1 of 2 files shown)


oscerd (approved):

Croway:
Hi @tadayosi , `jbang -Dhawtio.jbang.version=4.0.0.redhat-00029 hawtio@hawtio/hawtio` this should work as expected, but the productized version cannot be (easily) installed via `jbang app install ...` do you think the app install use case should be handled too?

Croway (approved):

tadayosi:
@Croway Thanks, you're right. So I asked JBang team and it appears `jbang app install` doesn't support such a use case yet: https://github.com/jbangdev/jbang/discussions/1736

So right now, what we could do are either using alias instead of `jbang app install`:
```
$ alias hawtio='jbang -Dhawtio.jbang.version=4.0.0.redhat-00029 hawtio@hawtio/hawtio'
```
or doing `jbang app install` but post-editing the installed script:
```
$ jbang app install hawtio@hawtio/hawtio
$ vi .jbang/bin/hawtio
#!/bin/sh
exec jbang run -Dhawtio.jbang.version=4.0.0.redhat-00029 hawtio@hawtio/hawtio "$@"
```

PR #3187, opened 2024-01-23
WIP: SAST fixes

mmelko:


hawtio-system/src/main/java/io/hawt/util/introspect/support/ClassScanner.java
tadayosi, line 1: It seems OK to me. Let's not just comment out but remove them now.

tadayosi (approved):
One comment but otherwise it looks good to me.

tadayosi:
@mmelko Also don't forget to change the title of the PR removing `WIP:` from it when you really think it's ready.

hawtio-system/src/main/java/io/hawt/util/introspect/support/ClassScanner.java
mmelko: done

PR #3188, opened 2024-01-25
fix(e2e): Update connect page test

kariuwu:


mmuzikar (changes requested):

tests/hawtio-test-suite/src/main/java/io/hawt/tests/features/pageobjects/pages/ConnectPage.java
```diff
@@ -24,7 +19,15 @@ public class ConnectPage extends HawtioPage {
...
     private static final By CONNECTION_LIST = By.id("connection-list");
 
+    private static final By CONNECTION_LOGIN_FORM = By.id("connect-login-form");
+
+    private static final By FOOTER_BUTTON = By.cssSelector("footer button.pf-m-primary");
+
     public void addConnection(String name, URL connection) {
+
+    if ($(CONNECTION_LIST).isDisplayed()) {
+        return;
```
mmuzikar, line 29: We should eventually double check that the connection name already exists and not expect that there will always be just one connection, you can just add a todo and explanation of why this workaround exists.

mmuzikar (changes requested):

tadayosi (approved):
Looks good to me other than what Martin pointed out.

mmuzikar (approved):

PR #3195, opened 2024-01-29
fix(quarkus): fix loadFromHawtioStatic sanitisation

tadayosi:
@mmelko I found a small issue with your previous fix. Please check.

The previous fix tried to sanitise the path for /hawtconfig.json but the path passed to loadFromHawtioStatic() is always constant thus forcing sanitising it is redundant without no real gain. We should instead clarify in the code that the path is always '/hawtconfig.json' to make the analyzer tool satisfy.

The previous fix also caused the following warning in the log:
```
Couldn't find file on location: file:///hawtconfig.json
```


oscerd (approved):

mmelko (approved):
LGTM

PR #3197, opened 2024-01-29
fix(ProxyServlet): Refactor how url being created when 300-303 status is received

mmelko:
I rewrote how URL is crafted. This should prevent open redirect vulnerability.  This code is reachable only when status from the proxy is between 300 and 303 (when redirect is necessary) which is not  often AFAIK.

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/proxy/ProxyServlet.java
+++ b/hawtio-system/src/main/java/io/hawt/web/proxy/ProxyServlet.java
@@ -341,7 +341,7 @@ public class ProxyServlet extends HttpServlet {
                     + " but no " + HttpHeaders.LOCATION + " header was found in the response");
             }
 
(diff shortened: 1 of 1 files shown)


hawtio-system/src/main/java/io/hawt/web/proxy/ProxyServlet.java
```diff
@@ -450,16 +450,17 @@ protected void copyResponseEntity(HttpResponse proxyResponse, HttpServletRespons
      * For a redirect response from the target server, this translates {@code theUrl} to redirect to
      * and translates it to one the original client can use.
      */
-    protected String rewriteUrlFromResponse(HttpServletRequest servletRequest, String theUrl, String targetUri) {
+    protected String rewriteUrlFromResponse( HttpServletRequest servletRequest, String theUrl, String targetUri) {
```
tadayosi, line 453: Unnecessary space at `( HttpServletRequest`.

tadayosi (approved):

PR #3200, opened 2024-02-01
fix(HAWNG-5): Remove unsafe-eval and unsafe-inline from the CSP header

mmelko:
Looks like we don't need this for the nawtio-next and we are not using inline scrips. I also verified that CSP header is modified and hawtio is working with this settings.

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/filters/ContentSecurityPolicyFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/filters/ContentSecurityPolicyFilter.java
@@ -28,8 +28,8 @@ public class ContentSecurityPolicyFilter extends HttpHeaderFilter {
 
     private static final String POLICY_TEMPLATE =
         "default-src 'self'; " +
-            "script-src 'self' 'unsafe-inline' 'unsafe-eval' %s; " +
-            "style-src 'self' 'unsafe-inline'; " +
+            "script-src 'self' %s; " +
+            "style-src 'self'; " +
             "font-src 'self' data:; " +
             "img-src 'self' data:; " +
             "connect-src 'self' %s; " +
--- a/hawtio-system/src/test/java/io/hawt/web/filters/ContentSecurityPolicyFilterTest.java
+++ b/hawtio-system/src/test/java/io/hawt/web/filters/ContentSecurityPolicyFilterTest.java
@@ -52,8 +52,8 @@ public class ContentSecurityPolicyFilterTest {
         contentSecurityPolicyFilter.addHeaders(request, response);
         // then
         verify(response).addHeader("Content-Security-Policy",
-            "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' ; "
-                + "style-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src 'self' data:; "
+            "default-src 'self'; script-src 'self' ; "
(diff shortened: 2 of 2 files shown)


hawtio-system/src/main/java/io/hawt/web/filters/ContentSecurityPolicyFilter.java
tadayosi, line 1: I remember these were carefully chosen in a previous ENTESB ticket. Can you explain how/why they were needed in Hawtio 2.x?  If you can answer, then we can decide whether it's really become unnecessary for Hawtio 3.x.

tadayosi (commented):
Please answer my comment.

hawtio-system/src/main/java/io/hawt/web/filters/ContentSecurityPolicyFilter.java
mmelko: yeah, you're right it was required mainly for the using the keycloak. See https://github.com/keycloak/keycloak/issues/9553 and https://github.com/keycloak/keycloak/issues/16277 . 

Let me test this with the Keycloak although I suspect I know what will be the result. probably it can be a bit more strict then it was originally. At least we know that hawtio it self is working with this.

tadayosi: If keycloak integration is the only reason we need these, then one possible option would be to enable them only when keycloak integration is enabled and always disable for the rest.

mmelko: I tested also integration with the keycloak running on remote host it works without `unsafe-eval` and `unsafe-inline`. 
After digging with into the code it looks it was brought by just enabling the CSP (https://github.com/jboss-fuse/hawtio/pull/368/commits/fa92403507178c6e624dcef53acb288529ef4545 backported from jboss-fuse) but they the key part for the keycloak to work was https://github.com/hawtio/hawtio/commit/5b77deb1d5453b06eceff78d531e2f281e94e4c9 

So it's good to get rid of `unsafe-eval` and `unsafe-inline`  because it diminishes the purpose of having CSP enabled.

tadayosi: Thanks. It looks you're right. Keycloak seems working fine without those directives.

One last concern with removing them from the header is whether the plugin loading works fine without those directives. Could you try hosting an external plugin somewhere else than the application itself and confirm the plugin loading still works fine with the fix?
https://github.com/hawtio/hawtio/blob/9ef985f6766de436597347172b6f256d781eb0c6/examples/springboot-plugin/src/main/java/io/hawt/example/spring/boot/SampleSpringBootService.java#L20-L29

tadayosi: To put it in other words, does Webpack Module Federation work fine without `unsafe-eval` and `unsafe-inline`?

mmelko: Ah! great idea. Will do

grgrzybek: Good point @tadayosi, I didn't think about it.

PR #3230, opened 2024-02-12
refactor(spring-authentication-quickstart): Update quickstart to work…

mmelko:
… with jetty version provided by spring-boot


oscerd (approved):

tadayosi (approved):
What's the reason behind this refactoring?  Why do we need to use SB's Jetty version?

mmelko:
Just that SB3.2 now officially supports jetty 12 (and servlet api 6). And it would be nice to align the quickstart dependencies with  the dependencies  from the SB starter.

PR #3235, opened 2024-02-20
4.x improvements

grgrzybek:


oscerd (approved):

hawtio-system/src/test/java/io/hawt/util/StringsTest.java
```diff
@@ -43,6 +44,56 @@ public void split() {
...
+            assertThat(Strings.resolvePlaceholders("${doubleProperty}"), is("hawtio.what"));
+            assertThat(Strings.resolvePlaceholders("${${doubleProperty}}"), is("world"));
+            assertThat(Strings.resolvePlaceholders("${${${doubleProperty}}}"), is("${world}"));
+        }
+
+        @Test
+        public void recursiveSystemProperties() {
+            System.setProperty("nestedProperty", "what");
+            System.setProperty("hawtio.what", "hawtio.${nestedProperty}");
+            assertThat(Strings.resolvePlaceholders("Hello ${hawtio.${nestedProperty}}!"), is("Hello hawtio.${nestedProperty}!"));
```
tadayosi, line 91: why isn't it resolved to `Hello hawtio.what!`?

hawtio-system/src/test/java/io/hawt/util/StringsTest.java
```diff
@@ -43,6 +44,56 @@ public void split() {
...
+            assertThat(Strings.resolvePlaceholders("Hello"), is("Hello"));
+            assertThat(Strings.resolvePlaceholders("Hello ${hawtio.what}!"), is("Hello world!"));
+            assertThat(Strings.resolvePlaceholders("${hawtio.what}"), is("world"));
+            assertThat(Strings.resolvePlaceholders("${hawtio.what}, hello ${hawtio.what}!"), is("world, hello world!"));
+        }
+
+        @Test
+        public void valuesWithPlaceholdersAreNotResolved() {
+            System.setProperty("hawtio.what", "${world}");
+            assertThat(Strings.resolvePlaceholders("Hello ${hawtio.what}!"), is("Hello ${world}!"));
```
tadayosi, line 63: there seems to be no test that verifies `Hello ${world}` resolves to `Hello ${world}` (a case when the placeholder system property doesn't exist in the first place).

hawtio-system/src/test/java/io/hawt/util/StringsTest.java
```diff
@@ -43,6 +44,56 @@ public void split() {
...
+        @Test
+        public void recursiveSystemProperties() {
+            System.setProperty("nestedProperty", "what");
+            System.setProperty("hawtio.what", "hawtio.${nestedProperty}");
+            assertThat(Strings.resolvePlaceholders("Hello ${hawtio.${nestedProperty}}!"), is("Hello hawtio.${nestedProperty}!"));
+            // no risk, because we don't resolve properties in system property values at all
+//            assertThrows(IllegalArgumentException.class,
+//                    () -> Strings.resolvePlaceholders("Hello ${hawtio.${nestedProperty}}!"));
+        }
+    }
```
tadayosi, line 96: I'd also check whether multibytes texts are processed without issues, such as:
```
System.setProperty("hawtio.what", "世界");
System.setProperty("こんにちは", "Hello");
"こんにちは、${hawtio.what}！" -> "こんにちは、世界！"
"${こんにちは} world!" -> "Hello world!"
```

tests/hawtio-test-suite/pom.xml
```diff
@@ -22,6 +22,21 @@
...
+        <artifactId>httpcore</artifactId>
+        <version>${httpcore-version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.httpcomponents</groupId>
+        <artifactId>httpclient</artifactId>
+        <version>${httpclient-version}</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
tadayosi, line 38: what's the motivation behind this?

tadayosi (approved):
LGTM.

I'd like to know the context of the changes. What are they supposed to be used for?

grgrzybek:
> LGTM.
> 
> I'd like to know the context of the changes. What are they supposed to be used for?

new `Strings` method will be used by OIDC part to load keystores/truststores for server-side http-client connections made (on initialization and cache invalidation) by OidcLoginModule to JWKS endpoint (to fetch public keys). I'd like to have everything configurable in single properties file where property values can have placeholders.

hawtio-system/src/main/java/io/hawt/util/Strings.java
```diff
@@ -80,4 +80,74 @@ public static String webContextPath(final String first, final String... more) {
         return cleanedPath.length() == 1 ? "" : cleanedPath;
     }
 
+    /**
+     * Simple, recursively-safe property placeholder resolver. Only system properties are used (for now). De-facto
+     * standard {@code ${...}} syntax is used. Unresolvable properties are not replaces and separators pass to
```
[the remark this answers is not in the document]
grgrzybek: my mistake - good catch ;)

hawtio-system/src/test/java/io/hawt/util/StringsTest.java
grgrzybek: First, nested `${nestedProperty}` is resolved to `what`, but because it's resolved inside other pair of `${...}` the value is another key, so we have `Hello ${hawtio.what}!`, but the value of `hawtio.what` system property (which is literally `hawtio.${nestedProperty}`) is not resolved recursively (we can do it if needed), but simply added to the output.

tests/hawtio-test-suite/pom.xml
grgrzybek: Without this, we have different versions of Http Client 4 visible in IDE (debugging problems):
image

Adding managed dependencies gives me:
image

Wrong (old) versions of httpclient/httpcore come from:
```
+- com.github.docker-java:docker-java:jar:3.3.4:compile
|  +- com.github.docker-java:docker-java-core:jar:3.3.4:compile
...
|  +- com.github.docker-java:docker-java-transport-jersey:jar:3.3.4:compile
...
|  |  +- org.apache.httpcomponents:httpcore:jar:4.4.13:compile
|  |  +- org.apache.httpcomponents:httpclient:jar:4.5.12:compile
```

PR #3326, opened 2024-03-25
[UI Fixes] Refactor E2E tests

jsolovjo:
Refactor Attributes details according to changes in https://github.com/hawtio/hawtio-next/pull/818

PR #3332, opened 2024-03-27
feat: update @hawtio/react to 1.2.0

tadayosi:


tadayosi:
The failing tests should be fixed with #3326.

PR #3365, opened 2024-04-02
build(console): migrate from craco to webpack

tadayosi:
hawtio/hawtio-next#455


tadayosi:
This refactoring seems stable enough. Let's merge it.

Issue #3178, opened 2024-01-11
[hawtio 3.0.0] Jolokia login incorrectly going through LDAP after losing connection, generating many errors due to refresh rate


amanda45762:
I have setup hawtio login to use LDAP using a login.conf file containing:

```
hawtio {
    com.sun.security.auth.module.LdapLoginModule REQUIRED
    userProvider="${LDAP_USER_PROVIDER}"
    authIdentity="{USERNAME}@****.com"
    userFilter="${LDAP_USER_FILTER}"
    useSSL=false
    debug=true;
};
```
These are my JVM args: ` -Dhawtio.authenticationEnabled: "true" -Dhawtio.authenticationContainerDiscoveryClasses: "" -Dhawtio.realm: "hawtio" -Djava.security.auth.login.config: "/opt/tomcat/conf/login.conf" -Dhawtio.sessionTimeout: "1200"  -Dhawtio.proxyAllowlist: "*" -Dendpoints.jolokia.sensitive: "false"
`
Hawtio login works fine. After logging into hawtio, I go to the connect tab and create a connection to my java app. When I click connect, theres a popup that prompts me for user/password creds that I've specified in the configuration options: `agentContext=/jolokia,user=***,password=***`. Once I've created a JMX connection, a **new** window automatically pops up where I can see the MBeans.

The bug occurs when the **original** window loses connection (or I logout of it), but the second window remains open and keeps retrying connections. These connection retries are going through LDAP, which is not configured for the jvm agent, so **the bug is that it is attempting to go through LDAP**. I am getting these LDAP login errors every 5 seconds (the default refresh rate):
`Login failed due to: Cannot bind to LDAP server   [LdapLoginModule] authentication-first mode; SSL disabled   [LdapLoginModule] user provider: ldap://****   [LdapLoginModule] attempting to authenticate user: ****   [LdapLoginModule] authentication failed   [LdapLoginModule] aborted authentication   `

I believe it's this section of hawtio code: https://github.com/hawtio/hawtio-next/blob/65b835cc58edcaa4728b575591766a9f71138665/packages/hawtio/src/plugins/shared/jolokia-service.ts#L344

My goal is to stop the connection retries from going through LDAP every 5 seconds without increasing session timeout or manually changing the refresh rate. Is there way to prevent the jolokia login retries and just logout the user entirely in all windows after session expiration? Or a way for the retries to not go through LDAP? Am I doing something wrong with the hawtio authentication realm or any other system properties I can use?

[...]

PR #3366, opened 2024-04-02
fix(#3178): Treat /proxy as /jolokia path and do not redirect to /aut…

grgrzybek:
…h/login

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationConfiguration.java
+++ b/hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationConfiguration.java
@@ -28,7 +28,7 @@ public class AuthenticationConfiguration {
     public static final String[] UNSECURED_PATHS = {
         "/login", "/auth/login", "/auth/logout", "/auth/config", "/auth/config/session-timeout",
         "/css", "/fonts", "/img", "/js", "/static", "/hawtconfig.json",
-        "/jolokia", "/user", "/keycloak", "/plugin"
+        "/jolokia", "/proxy", "/user", "/keycloak", "/plugin"
     };
 
     // =========================================================================


oscerd (approved):

grgrzybek:
@tadayosi please decide whether to merge it. IMO, `/jolokia` and `/proxy` should be treated equally by `io.hawt.web.auth.LoginRedirectFilter` - this is how it worked with Hawtio 1.

tadayosi:
Note `ProxyServlet` can be easily a target for security attacks such as CSRF, so we'd need to be careful to make it public.

grgrzybek:
> Note `ProxyServlet` can be easily a target for security attacks such as CSRF, so we'd need to be careful to make it public.

Good point and actually a clarification of our discussion.

Let me summarize the discussion here:
* there's _main_ Hawtio tab and _connect_ tabs (for remote Jolokia agents accessed via proxy).
* for _connect_ tabs, `/proxy` requests are being sent with `Authorization` header specific to remote Jolokia, not to Hawtio itself (that's why `io.hawt.web.auth.LoginRedirectFilter#tryAuthenticateRequest()` from 243393102a44c01bd7eb97d31875b43cb45c484a should NOT be called for proxied requests)

We were wondering whether logging out (explicitly or by session timeout) from _main_ Hawtio tab should lead to logout of all _connect_ tabs.
* without this PR, after _main_ tab logs out, _connect_ tabs keep sending `/proxy` requests and are redirected to `/auth/login` AND also the credentials for remote Jolokia are used for Hawtio login (which is wrong). This leads to hawtio/hawtio-next#705 - "Connection lost. Retrying..." is being shown. The point is that it's a connection to proxy servlet, not to the remote agent. That's why I added `/proxy` to the list here.

However, if we explicitly admit that _`ProxyServlet` can be easily a target for security attacks such as CSRF, so we'd need to be careful to make it public_, we should rather put `/proxy` together with `/jolokia` under the mapping for `AuthenticationFilter`... This will cause 403 from AuthenticationFilter before 302 from LoginRedirectFilter... Let me check this.

tadayosi:
Thinking about it a bit further, now I'm not sure if this change is really the right direction. The change is intended to expose an originally protected endpoint for an edge case hawtio/hawtio-next#705, but it also means it exposes the endpoint in the normal cases, which has been secure, and anyone can access it. In terms of impact, this will open an unnecessary hole in the majority of normal cases in order to support a rare case.

If we really need to expose the proxy, we might as well need to add some other way to maintain the same level of security in the normal cases.

grgrzybek:
Seeing all the risks, I'd conclude with a statement:

> Hawtio's ProxyServlet is a feature that should be protected.

More precisely:
* even if the remote Jolokia agent doesn't maintain session, session timeout or explicit logout from Hawtio should forbid user from using Hawtio's proxy servlet and _connect_ tab should be logged out too
* we can't simply map `/proxy/*` to AuthenticationFilter, because we actually can't authenticated the user - because `Authorization` header is provided for remote Jolokia, not for us (Hawtio)
* LoginRedirectFilter should NOT redirect unauthenticated `/proxy/*` request to `/auth/login`. 403 should be returned instead (but by different filter or ProxyServlet itself) - this will ensure that _connect_ tab is logged out.

grgrzybek:
Of course it's more complicated with Keycloak/OIDC, because there's no "subject" session attribute set by LoginServlet...

grgrzybek:
Closing this PR, as this needs to be solved differently (see: #3178, #3371, #3379)

PR #3370, opened 2024-04-04
[E2E Tests] Refactor Chart tests according to UX changes

jsolovjo:
According to UX changes in Charts, there is a test refactoring.

New UX changes also brought a few difficulties as chart's title is not as one complete text, but several texts, so a more complex solution with Xpath has been applied where it's needed to select the desired title using text's index.


tadayosi (approved):

tadayosi:
@jsolovjo Let's merge it when we are ready to port @mmelko's chart refactoring to hawtio/hawtio.

mmelko (approved):

PR #3392, opened 2024-04-11
[E2E Tests] Add sorting attributes tests

jsolovjo:
- Add new tests for sortable attributes


jsolovjo:
The checks failed because hawtio and hawtio-next are not synced yet as I understand that

tadayosi (approved):

PR #3397, opened 2024-04-17
fix(connect): Handle redirect with "con" parameter for proxy requests

### LINKED ISSUE #3317: Connecting to a remote JVM opens the same tab instead
Kinda struggling the last couple days trying to pin an issue with remote connections after upgrading from Spring Boot 2.7 to Spring Boot 3.2.

I have the following setup here:

- Spring Boot Admin Server with Hawt.io console attached
- Several applications with Apache Camel
- Spring Actuator for application monitoring on all apps
- Jolokia enabled through Actuator on all apps

The main goal is to have a central administrative interface for Apache Camel using a single hosted Hawt.io console that connects to the remote apps.

In the Hawt.io console I have set up the connections to the remote agents manually, testing the connection works just fine.
On my Spring Boot 2.7 stack  with Hawt.io 3 opening a remote connection opens a new tab with a console showing the data from the remote agent.
On the Spring Boot 3.2 stack with Hawt.io 4 however the new opened tab shows the same local console. 

I found out that if I switch to a different plugin like JMX on the local console then add the 'con' parameter to the browser URL I now get the remote console of the connection provided. I can then proceed to manage the remote JVM like normal including Camel which is missing on the local console.

The current stack as follows:
- Spring Boot 3.2.0
- Hawt.io using hawtio-springboot 4.0.0-RC1
- jolokia-support-spring 2.0.1 on remote apps via Spring Actuator

Is there anything I'm missing here?

Regards

grgrzybek:
When opening new connection tab from Hawtio on Spring Boot, we're getting redirected without con parameter. This commit fixes this problem.

Fixes #3317

### DIFF
--- a/platforms/springboot/src/main/java/io/hawt/springboot/HawtioEndpoint.java
+++ b/platforms/springboot/src/main/java/io/hawt/springboot/HawtioEndpoint.java
@@ -43,6 +43,10 @@ public class HawtioEndpoint implements WebMvcConfigurer {
         final String path = endpointPath.resolve("hawtio");
 
         if (request.getRequestURI().equals(path)) {
+            String query = request.getQueryString();
+            if (query != null && !query.isEmpty()) {
+                return "redirect:" + path + "/index.html?" + query;
+            }
             return "redirect:" + path + "/index.html";
         }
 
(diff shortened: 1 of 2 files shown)


tadayosi (approved):
:+1:

tadayosi:
Thank you!

PR #3408, opened 2024-04-24
fix: Upgrades springboot to 3.2.5

phantomjinx:
* The CVE requires the upgrade of spring products to 6.1.6+. To achieve this upgrade both the spring version and the springboot version in the pom configuration.

* See CVE described by https://spring.io/security/cve-2024-22262


tadayosi (approved):
Note that we cannot yet upgrade 4.x-redhat branch until CSB is upgraded to a version that supports those SB versions.

PR #3410, opened 2024-04-25
e2e: rerun failed tests

mmuzikar:
This will rerun failed tests up to 3 times, the tests that pass on the retry are marked as flakes. For example: sometimes it happens that Hawtio doesn't load in time (there's about 30 second wait time), while we could always increase the wait time, it makes more sense to rerun the test again with the sensible time.


tadayosi (approved):

tadayosi:
thank you!

PR #3440, opened 2024-05-03
Fix RP attachments for Cucumber

mmuzikar:
Fixes attachments to RP for Cucumber tests, before this change RP reports didn't contain the screenshot proving quite useless in terms of a report.


mmelko (approved):
LGTM

PR #3442, opened 2024-05-07
[HAWNG-641] Set text/plain content-type for "refresh" response. No ne…

grgrzybek:
…ed for Content-Security-Policy header


tadayosi (approved):
thx!

oscerd (approved):

PR #3449, opened 2024-05-09
e2e: add camel-k integration to Hawtio Online tests

mmuzikar:
Allows using a Camel-K integration for E2E tests. Normally the latest Red Hat version from OperatorHub is used, but it can also be configured to use a different catalog by a property. 
Updated the readme to mention Camel K as well as changing / removing some outdated info. IMO it's not worth it to state the versions in the README as it's not always up to date with the pom

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -2,18 +2,11 @@
 
 ## Introduction
 
-E2E UI tests
-
 ### Test tools and frameworks
 
 - Selenium 4.x
 - Selenide 6.x
 - Cucumber 7.x
-
-### Runtime Support
-
-- Spring Boot 3.x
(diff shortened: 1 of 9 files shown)


tests/README.md
```diff
@@ -3,18 +3,6 @@
 ## Introduction
 
 E2E UI tests
-
-### Test tools and frameworks
-
-- Selenium 4.x
-- Selenide 6.x
-- Cucumber 7.x
```
tadayosi, line 11: its ok to drop versions but I think its worth mentioning what testing frameworks are used/required for new contributors.

tadayosi (approved):

PR #3452, opened 2024-05-13
[E2E Tests] Disable default HTTPS scheme when creating a new connection

jsolovjo:
The changes introduced in https://github.com/hawtio/hawtio-next/pull/930 cause E2E tests failures.

It happened due to default HTPPS scheme when a new connection is created.

The fix for E2E tests is to switch from default HTTPS to HTTP scheme when creating a new connection.


phantomjinx (commented):
LGTM

phantomjinx (approved):

PR #3455, opened 2024-05-20
feat(connect): Authenticate /proxy/* requests with proper session han…

### LINKED ISSUE #3178: [hawtio 3.0.0] Jolokia login incorrectly going through LDAP after losing connection, generating many errors due to refresh rate

grgrzybek:
…dling (fixes #3178)


oscerd (approved):

hawtio-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -105,6 +105,10 @@
     <filter-name>AuthenticationFilter</filter-name>
     <url-pattern>/jolokia/*</url-pattern>
   </filter-mapping>
+  <filter-mapping>
+    <filter-name>AuthenticationFilter</filter-name>
+    <url-pattern>/proxy/*</url-pattern>
+  </filter-mapping>
```
tadayosi, line 111: `<url-pattern>` should be able to be defined multiple times in a single `<filter-mapping>`.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/filters/HawtioQuarkusAuthenticationFilter.java
```diff
@@ -31,33 +33,79 @@ public class HawtioQuarkusAuthenticationFilter extends AuthenticationFilter {
...
     private static final Logger LOG = LoggerFactory.getLogger(HawtioQuarkusAuthenticationFilter.class);
 
+    private InstanceHandle<HawtioQuarkusAuthenticator> handle;
     private HawtioQuarkusAuthenticator authenticator;
 
     @Override
     public void init(FilterConfig filterConfig) throws ServletException {
-        authenticator = Arc.container().instance(HawtioQuarkusAuthenticator.class).get();
+        handle = Arc.container().instance(HawtioQuarkusAuthenticator.class);
+        authenticator = handle.get();
```
tadayosi, line 42: I think then this `handle` should be renamed to `authenticator` and the original raw `authenticator` should be removed in favour of always doing `authenticator.get()` when needed.

tadayosi (approved):
Two minor comments but overall it looks good to me.

hawtio-war/src/main/webapp/WEB-INF/web.xml
grgrzybek: oh my - I forgot ;)

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/filters/HawtioQuarkusAuthenticationFilter.java
tadayosi: Now I don't think it's a good idea. Your original one is better. Let me close it.

grgrzybek: You forgot we're not in OSGi ;)

PR #3517, opened 2024-06-10
fix(system): monaco-editor src version should be wildcarded in CSP header

tadayosi:


tadayosi:
Ah, wildcards cannot be used in the path part. So we need to make sure to update the policy per monaco-editor upgrade.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources
> Wildcards ('*') can be used for subdomains, host address, and port number, indicating that all legal values of each are valid.

Closing.

PR #3533, opened 2024-06-20
Remove redundant camel-quarkus-jaxb from Quarkus example project

jamesnetherton:
Minor UX improvement.

`camel-quarkus-management` depends on `camel-quarkus-xml-jaxb`, which is enough for route dumping etc. In future CQ releases, the management extension will use the lightweight xml-io component, so it's probably best to clean up the example ahead of this.

### DIFF
--- a/examples/quarkus/pom.xml
+++ b/examples/quarkus/pom.xml
@@ -72,12 +72,6 @@
       <artifactId>camel-quarkus-management</artifactId>
     </dependency>
 
-    <!-- To enable Camel route dumping to XML. Required for the route diagram -->
-    <dependency>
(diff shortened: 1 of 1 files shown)


tadayosi (approved):
thanks!

tadayosi:
@jamesnetherton I remember last time I tried xml-io the xml dumping didn't work. So can we expect it to work in an updated version?

jamesnetherton:
> Can we expect it to work in an updated version?

Yes, from the CQ 3.13.0 release (next month). We have tests for route dumping, so we'll know whether it works or not.

I'll test HawtIO with the latest CQ SNAPSHOT before we release to be sure it works.

PR #3598, opened 2024-08-28
fix(HAWNG-751): Webpack upgrade

mmelko:
Upgrade webpack to 5.94.0


tadayosi (commented):
Why don't we do upstream first with `4.x` branch?

tadayosi (changes requested):
Also, don't forget to update `examples/sample-plugin` too.

tadayosi (approved):

PR #3623, opened 2024-09-12
fix(HAWNG-770): Upgrade to Jolokia 2.1.1 (with own JSON parser)

grgrzybek:


oscerd (approved):

grgrzybek:
@tadayosi I think we'd need @hawtio/react release first...
Though this change is only on Java side, not JavaScript side of hawtio/hawtio...

tadayosi (approved):

tadayosi:
@grgrzybek All the e2e failed. You can run e2e locally to see what went wrong before merging the fix.

PR #3626, opened 2024-09-16
fix(HAWNG-802): Upgrade to Spring Framework 6.1.13

grgrzybek:


grgrzybek:
FYI - I've added Spring Framework BOM before Spring Boot BOM in hawtio examples, so the versions from SPring Framework override Spring Boot versions (which may for a while lag behind).

PR #3655, opened 2024-10-08
fix: HAWNG-851 Update RHBAC for Spring Boot version to 4.8.0

tadayosi:
With this, upstream E2E will no longer pass green until RHBAC for SB 4.8.0 version is released.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,12 +55,12 @@
     <quarkus-version>3.8.4.SP1-redhat-00001</quarkus-version>
     <quarkus-extension-version>3.8.4.redhat-00002</quarkus-extension-version>
     <!-- Spring -->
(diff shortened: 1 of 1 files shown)


tadayosi:
cc @grgrzybek

grgrzybek:
So upstream (`4.x` branch) needs upstream CSB 4.8.0 - do I understand correctly?

tadayosi:
@grgrzybek upstream `4.x` already uses CSB 4.8.0, but I don't think we need to keep upstream and the product branch in sync for CSB versions. Once CSB 4.9.0 is released we can upgrade to it for upstream Hawtio 4.x.

tadayosi:
E2E for SB passed locally.

PR #3671, opened 2024-10-25
fix(monaco): Do not add Monaco CDN location to CSP header (Monaco is …

grgrzybek:
…embedded in @hawtio/react 1.6.1+)

See:
* hawtio/hawtio-next#1186


tadayosi (changes requested):
Tests also need to be changed.

And it should be waited for merge until we publish hawtio/react 1.6.1 and update it for hawtio/hawtio.

PR #3710, opened 2024-11-20
e2e: add cleanup after throttling tests

mmuzikar:


tadayosi (approved):
Looks good, but also note that throttling state is cached at the server side. We might need to restart the server runtime, or isolate the test scenario from the rest.

PR #3789, opened 2025-02-03
e2e: add automated testing support for keycloak on quarkus

mmuzikar:
See passing runs: https://github.com/mmuzikar/hawtio/actions/runs/13110882131
Runs the standard testsuite using admin account on keycloak.


tadayosi (approved):
GJ!

PR #3795, opened 2025-02-06
Jetty auth

grgrzybek:
Generally I've removed the PropertyFileLoginModule from examples/springboot-authentication and created special `platforms/jetty-security` Maven module, which:
* contains `io.hawt.jetty.security.jaas.PropertyFileLoginModule` extending `org.eclipse.jetty.security.jaas.spi.PropertyFileLoginModule` only to make it work in more scenarios (fixing jetty/jetty.project#6223) without reimplementing the logic
* contains ready to use [Jetty Module Definition](https://jetty.org/docs/jetty/12/operations-guide/modules/index.html) which I've documented in hawtio/website#64


hawtio-system/src/main/java/io/hawt/web/auth/LogoutServlet.java
```diff
@@ -35,7 +35,10 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
         // Send some HTTP headers on logout
         addHeaders(response);
 
-        request.logout();
+        if (request.getUserPrincipal() != null) {
+            request.logout();
+        }
```
tadayosi, line 40: Why is this necessary?  What happens if the request doesn't provide a principal but still firing logout is necessary?

tadayosi: https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#logout--

tadayosi (approved):

hawtio-system/src/main/java/io/hawt/web/auth/LogoutServlet.java
grgrzybek: I agree - the doc suggests that it _should be_ a no-op if there's no request-bound principal.
But with Jetty we're getting:
```
jakarta.servlet.ServletException: logout failed
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.logout(ServletApiRequest.java:626)
	at jakarta.servlet.http.HttpServletRequestWrapper.logout(HttpServletRequestWrapper.java:293)
	at io.hawt.web.auth.LogoutServlet.doGet(LogoutServlet.java:38)
...
```

```java
@Override
public void logout() throws ServletException
{
    ServletRequestInfo servletRequestInfo = getServletRequestInfo();
    if (!AuthenticationState.logout(getRequest(), servletRequestInfo.getServletChannel().getServletContextResponse()))
        throw new ServletException("logout failed");
}
```

and `org.eclipse.jetty.security.AuthenticationState#logout()` returns `false` if there's no principal.

The thing is that Hawtio is _not_ using `<login-config />` section in `web.xml`...

grgrzybek: Tomcat has better code:
```java
Principal p = request.getPrincipal();
if (p instanceof TomcatPrincipal) {
    try {
        ((TomcatPrincipal) p).logout();
    } catch (Throwable t) {
        ExceptionUtils.handleThrowable(t);
        log.debug(sm.getString("authenticator.tomcatPrincipalLogoutFail"), t);
    }
}
```

PR #3811, opened 2025-02-28
fix(quarkus): Set HttpOnly flag on JSESSIONID cookie for hawtio-quarkus (HAWNG-945)

grgrzybek:
@jamesnetherton may I ask you to check if I did it the right "Quarkus" way?

### DIFF
--- a/platforms/quarkus/deployment/src/main/java/io/hawt/quarkus/deployment/HawtioProcessor.java
+++ b/platforms/quarkus/deployment/src/main/java/io/hawt/quarkus/deployment/HawtioProcessor.java
@@ -12,6 +12,7 @@ import io.hawt.quarkus.HawtioConfig;
 import io.hawt.quarkus.HawtioProducers;
 import io.hawt.quarkus.HawtioRecorder;
 import io.hawt.quarkus.auth.HawtioQuarkusAuthenticator;
+import io.hawt.quarkus.auth.SessionConfigurationServletExtension;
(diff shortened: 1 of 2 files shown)


jamesnetherton:
It's probably ok. I think this is the only way for Hawtio to set this up.

If folks were to include their own Servlet in their app along with a custom `web.xml` that contained `<session-config>`. Then I am not sure which config would 'win'.

That is likely to be something of an edge case though.

grgrzybek:
@jamesnetherton actually Hawtio is reading web.xml which includes:
```xml
<session-config>
  <cookie-config>
    <http-only>true</http-only>
  </cookie-config>
</session-config>
```

But what it does is: https://github.com/hawtio/hawtio/blob/43b5a4fdfe29da5e91fc9da05087d6c1e9817c96/platforms/quarkus/deployment/src/main/java/io/hawt/quarkus/deployment/HawtioProcessor.java#L108-L115

So manual reading and taking servlets, filters and listeners only...

jamesnetherton:
> @jamesnetherton actually Hawtio is reading web.xml

Yes, I wrote that code 🙂. But the Hawtio `web.xml` is processed outside the scope of the user application.

In theory a user could provide their own `META-INF/resources/web.xml` containing its own `session-config` and the `quarkus-undertow` extension would read it. 

But, like I said, it's a bit of an edge case TBH. Probably not worth investing time in investigating.

jamesnetherton (approved):

PR #3833, opened 2025-03-04
chore(build): Attach Jetty hawtio-jaas.mod in dedicated profile

grgrzybek:
With this profile, `hawtio-jaas.mod` will be attached to the build normally, but there'll be an option to disable this attachment with `-P-jetty-module`.


phantomjinx (approved):
LGTM

PR #3842, opened 2025-03-17
[E2E] Fix failing tests in PR for issue 1378

jsolovjo:
This PR is to fix the tests in https://github.com/hawtio/hawtio-next/pull/1405 and includes the following:
- Removed a feature file covering tests in Camel Contexts as now it doesn't make sense as the first page on Camel now shows the routes and there is only one Camel Context page;
- Added a few extra steps to explicitly open Routes tab;
- A small workaround added to explicitly navigate back to Routes tab after an operation is executed in Routes tab as it automatically redirects to Route Diagram tab;
- Another small issue is when a new endpoint is added from URI or Data, the UI doesn't show it automatically, you have to either refresh the page or change a view, e.g. go to another page and return back. The workaround is to click on the same endpoints folder and it will show the endpoint;


grgrzybek (approved):

mmuzikar (approved):

tadayosi (approved):
Awesome, thank you!

PR #3874, opened 2025-04-17
Remove deprecated Pragma header

GChuf:
Pragma header is deprecated - Cache-Control is enough.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma


tadayosi (commented):
Thanks! Looks good but we need to check internally that removing the header is really ok for our security best practice.

GChuf:
Sorry for the mess - force pushed to fix author name.

PR #3883, opened 2025-04-29
chore(deps): Upgrade to Jetty 12 (fixes #3794)

### LINKED ISSUE #3794: Upgrade hawtio-embedded to Jetty 12
Jetty 10 and 11 [are no longer supported](https://github.com/jetty/jetty.project/issues/10485) and Jetty 12 should be the one and only version used (supporting javax/jakarta packages and JakartaEE 8+ versions).

Not a short term goal, but eventually it'd be nice to get rid of previous Jetty versions.
Spring Boot 3.4.2 BOM uses Jetty 12.0.16.

cc: @tadayosi

grgrzybek:


deploy/hawtio-embedded/pom.xml
```diff
@@ -25,14 +25,26 @@
...
+      <groupId>io.hawt</groupId>
+      <artifactId>hawtio-system</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.hawt</groupId>
+      <artifactId>hawtio-local-jvm-mbean</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
```
tadayosi, line 37: are these `test` intentional?  why do we introduce them?

tadayosi (approved):
Thanks! LGTM. Just one minor question.

deploy/hawtio-embedded/pom.xml
grgrzybek: without these, `io.hawt.embedded.MainTest` doesn't work in IDE and works only with `mvn clean install`, because it requires packaged jar artifact. `test` scope is safe - I checked that it doesn't affect the target artifact.

GChuf:
@tadayosi @grgrzybek  jetty12 requires java17 - does it make sense to bump compiler version for hawtio to 17, and drop support for java 11 ... ?

grgrzybek:
Same is for Spring Boot 3.x... So +1 from me.

tadayosi:
@grgrzybek I just wonder if Artemis team are happy with our dropping Java 11 support?  AFAIR Artemis still supports 11, doesn't it?

GChuf:
Both activemq artemis and the artemis console now require java17+
https://github.com/apache/activemq-artemis/commit/755c40daea2269bf705dce6330dc92395798c621
https://github.com/apache/activemq-artemis-console/commit/83fb5aa617674547f7dc188e3cefc8fca7147e4c

PR #3918, opened 2025-05-22
chore(deps): update jetty servlet api

GChuf:
Bumps servlet-api, jetty 12 supports v6.x.
v5.0.0 was compiled with java 1.8, and v6.1.0 with java 11.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
     <mockito-core-version>5.16.1</mockito-core-version>
     <nimbusds-jose-version>10.0.2</nimbusds-jose-version>
     <picocli-version>4.7.7</picocli-version>
-    <servlet-api-version>5.0.0</servlet-api-version>
+    <servlet-api-version>6.1.0</servlet-api-version>
     <!-- Pin SLF4J version until https://issues.apache.org/jira/browse/CAMEL-20598 is resolved at SLF4J -->
     <slf4j-version>2.0.11</slf4j-version>
 


tadayosi (commented):
Jetty 12 still supports Jakarta EE 9. Not sure if it's the right direction to drop support for Jakarta EE 9 now, as it would drop support for deploying our wars to older web containers.
https://jetty.org/docs/jetty/12/index.html

@grgrzybek WDYT?

grgrzybek:
I checked that Hawtio correctly uses Servlet API dependency in all the places:
```xml
<dependency>
  <groupId>jakarta.servlet</groupId>
  <artifactId>jakarta.servlet-api</artifactId>
  <version>${servlet-api-version}</version>
  <scope>provided</scope>
</dependency>
```

scope=provided is how this API should be used.

It'd matter much more if we had OSGi support (we could get `Import-Package: jakarta.servlet; version="[6.0, 7.0)"` and it'd break on JakartaEE 9 / Servlet API 5).

However without OSGi it matters less, because dropping a WAR into Tomcat 10 would work without problems _unless_ Hawtio starts using some API which is added only in Servlet API 6 (like https://github.com/jakartaee/servlet/issues/407).

@GChuf I'd however treat this Hawtio _dependency_ as kind of baseline - "we're supporting JakartaEE 9++ / Servlet API 5++". This has nothing to do with Jetty 12 / Tomcat 11.

tadayosi:
> unless Hawtio starts using some API which is added only in Servlet API 6 (like https://github.com/jakartaee/servlet/issues/407).

Yeah, I think this is the most important part for keeping using the oldest supported APIs.

tadayosi:
So we don't want to upgrade Servlet API right now. Please let me close it for now because I don't want to merge it by accident.

PR #3948, opened 2025-06-16
[E2E] Camel Route Diagram tests to check overlaps and duplications.

jsolovjo:
In this PR:
- Added a test to check whether route nodes overlaps/intersect;
- Added a test to check for uniqueness (no duplications);

Those tests should cover regressions such as in the https://github.com/hawtio/hawtio-next/issues/1543


phantomjinx:
Nice work!

PR #4012, opened 2025-08-08
[E2E] Update snapshot versions.

jsolovjo:
Update snapshot version in tests


jsolovjo:
> i believe you should also update some other parts in tests such as dockerfile. lets grep it to check the places.

@tadayosi I updated the snapshot versions in more places, can you check it, please?
Thanks

PR #4066, opened 2025-09-09
Change a numer of the login attempts for the user lockout test

kariuwu:


tadayosi (commented):
Is this the root cause of the flakiness?  That sometimes the loading of the login page takes time and needs a warmup?

Isn't there any suspicion that the startup of authentication mechanism fails sometimes and then never gets ready?

kariuwu:
> Is this the root cause of the flakiness? That sometimes the loading of the login page takes time and needs a warmup?
> 
> Isn't there any suspicion that the startup of authentication mechanism fails sometimes and then never gets ready?

The starnge thing is that the lockout user case looks like it passess the first time (it locks the user for 1 second), but then it fails at 3 seconds mark. I am figuring out why.

tadayosi:
@dependabot rebase

PR #4068, opened 2025-09-11
Refactor Jbang test workflow

kariuwu:


kariuwu:
@tadayosi Most likely the checks will fail, because the workflows changes are not applied yet. What do you think; is it okay to merge it with failing checks or we should test it the way Martin did while developing the workflows?

tadayosi (approved):
yeah this is trivial so you can merge it right away

PR #4069, opened 2025-09-12
Remove obsolete HPKP http header

GChuf:
HPKP is now obsolete and "replaced" by CT:

https://developer.mozilla.org/en-US/docs/Glossary/HPKP
https://developer.mozilla.org/en-US/docs/Web/Security/Certificate_Transparency

Additional discussion here:
https://github.com/apache/activemq-artemis-console/pull/117


grgrzybek:
see:
* #4070

grgrzybek:
Thanks @GChuf !

Issue #2941, opened 2023-09-15
Support multiple authentication methods at runtime


tadayosi:
Users should be able to have an option to choose one of the multiple configured authentication methods at runtime, i.e. the hawtio login page could have a new button "Continue with Keycloak" and users could choose to login using the Hawtio login page (typing username/password) or to login using keycloak by clicking a Continue with Keycloak button.

grgrzybek:
Another observation. With OIDC (and actually with ancient `TomcatAuthenticationContainerDiscovery`) we use two distinct JAAS configurations:
* when doing OIDC, we create `javax.security.auth.login.LoginContext` instance using 4th argument where `Configuration` is passed
* when not doing OIDC (or Tomcat is not detected or Spring Security is not enabled) we don't use 4th argument and pass only realm name, which means the `javax.security.auth.login.Configuration#getConfiguration()` is used where the file is read using `sun.security.provider.ConfigFile.Spi` and `-Djava.security.auth.login.config` specified location

This needs to be more clever and we need to merge app configuration entries from static and dynamic locations.

PR #4103, opened 2025-10-17
Jolokia 2.4 update

grgrzybek:


grgrzybek:
@tadayosi I know this is a lot... I hope to never have to push such mixed commits in one PR.

The _big_ change is:
* #2941

And I had to do serious review of JAAS, Tomcat "special" Login Module and configuration servlets...

I'm not asking for review, because it'd take too long of your time - I tried to add as many comments as possible.

Additionally I'm upgrading to just released Jolokia 2.4.0 (for which I have to prepare documentation... but that's after the relase) and to new `@hawtio/react` which I also upgraded to `jolokia.js` 2.4.0.

This is all needed to finally merge the changes for JFR and focus on checking the authentication in Hawtio Online (cc @phantomjinx)

PR #4107, opened 2025-10-22
[E2E] Create auth enabled and disabled test app images and refactor t…

jsolovjo:
In this PR:

– Workflow file refactoring
– -The workflow image file now creates test application images with and without authorization (for online tests)
– Replace previously used test application images for online tests and configure them to be disabled for authorization and accessible via Hawtio Online
– Update the XTF library to avoid a fabric8/jakcson serialization error when patching/editing OpenShift resources
– Refactor tests to adapt to the updated test application to avoid flakiness and improve stability

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -31,23 +31,39 @@ jobs:
       - name: Install
         run: |
           mvn ${MVN_ARGS} install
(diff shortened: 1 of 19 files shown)


grgrzybek:
Great work @jsolovjo !

PR #4116, opened 2025-10-30
chore(build): Align deps and webpack.config.js with @hawtio/react 1.11.1

grgrzybek:


tadayosi (changes requested):
Please update console-minimal as well. We need it updated for releasing.

PR #4145, opened 2025-11-11
[E2E] Fix the image workflow and adjust Dockerfiles.

jsolovjo:
This PR fixes the issue related to messed paths and POM files in the Build Image workflow.

I triggered the workflow manually, it produced and pushed the images - https://github.com/hawtio/hawtio/actions/runs/19263938890

Then I checked them by deploying test applications on OpenShift 4.21 running on s390x and on classic x86_64

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -52,17 +52,11 @@ jobs:
           fi
           docker run --rm \
           -v /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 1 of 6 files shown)


tadayosi (approved):

jsolovjo:
Hi @tadayosi 
The e2e checks will fail here as they do not use the corrected workflow file from the PR.
Now the e2e should be fixed, I checked it in my forked project - https://github.com/jsolovjo/hawtio/actions/runs/19271034371

Issue #4155, opened 2025-11-22
Application property to disable Hawtio


robertonav20:
Hi,
at the moment it seems is not possible to disable hawtio beans with a property, if it's correct could be very useful disable it with a configuration

grgrzybek:
What do you exactly mean by _Hawtio beans_?

robertonav20:
I would refer to all hawtio suite, controller, log dashboard and so on?
Just another question an upgrade to spring boot 4.x.x is planned?

grgrzybek:
by "log dashboard" I could understand the log plugin (enabled when `hawtio:type=LogQuery` MBean is available).
but what do you mean by "hawtio suite" and "hawtio controller"?

robertonav20:
Hi @grgrzybek 
sorry i didn't describe well, i would have the chance to disable Hawtio just with a configuration like `hawtio.enabled`, so all bean created be these classes 

- HawtioEndpointAutoConfiguration
- HawtioPluginAutoConfiguration
- HawtioLogAutoConfiguration

I made some analysis and there isn't a way to disable all `Hawtio beans` with a property, is something which could be added?
Otherwise we i should exclude them in some way. 
Are you agree with my analysis or i miss something?

 I can create a PR to introduce it this property 😄

tadayosi:
Hi @robertonav20,

Why don't you simply exclude `hawtio-springboot` from your POM?
These MBeans are the backend services for some of the frontend plugins and features. Disabling only those MBeans would stop Hawtio work correctly.

robertonav20:
Hi @tadayosi 
because I would use it on local/dev environment only and disable it on upper environment.
Is a scenario compliant to hawtio plan?

Of course if I remove dependency on pom I will exclude it for all

robertonav20:
Hi,
i opened the PR https://github.com/hawtio/hawtio/pull/4157 to introduce the variable, now i see workflow not completed successfully because there is a Vulnerability related to https://www.npmjs.com/package/body-parser of hawtio-console.

Let me know if you are agree with me 😄

tadayosi:
Thank you for your contribution!

However, I'm still not convinced that it's really needed. It's totally fine that you enable Hawtio only on local/dev environments and disable it in production. We support it.
But I believe this can be done simply by using a profile in the POM and generating different builds between the local/dev and production envs.
Is it your goal that you want to switch the application's configuration only with the environment and application properties without changing the deployment? I'm open for discussions.

[...]

PR #4157, opened 2025-11-26
[#HWT4155] Configuration - added property to manage activation

robertonav20:
Hi
this is my PR to manage the scenario of https://github.com/hawtio/hawtio/issues/4155.

i added a new property `hawtio.enabled` to manage hawtio activation default is true

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ You can attach the Hawtio console to your Spring Boot app with the following ste
    <dependency>
      <groupId>io.hawt</groupId>
      <artifactId>hawtio-springboot</artifactId>
-     <version>4.6.1</version>
+     <version>4.6.2</version>
    </dependency>
    ```
 
@@ -89,7 +89,7 @@ mvn spring-boot:run
 
 Opening <http://localhost:8080/actuator/hawtio> should show the Hawtio console.
 
-See [Spring Boot example](https://github.com/hawtio/hawtio/tree/hawtio-4.6.1/examples/springboot) for a working example app.
+See [Spring Boot example](https://github.com/hawtio/hawtio/tree/hawtio-4.6.2/examples/springboot) for a working example app.
 
 ### Running a Quarkus app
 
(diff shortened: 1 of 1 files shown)


hawtio-jbang/dist/HawtioJBang.java
tadayosi: this shouldn't be changed

tadayosi:
please rebase it, I just fixed the body-parser issue.

grgrzybek:
So `matchIfMissing = true` is good - we shouldn't change the behavior if there's no `hawtio.enabled` in the properties.

As for other runtimes, I don't think we need this property for WAR deployments - you simply don't deploy Hawtio WAR if you don't need it ;)

For Quarkus, I think (being aware that Quarkus is more build-time oriented), I also don't think it's critical to have such property. You simply provide different build, where everything (including configuration) is kind of prebuilt by Quarkus Maven Plugin.

This flag seems much more fitted for Spring Boot deployments, because that's how the autoconfigurations work.

The only thing - can you @robertonav20 switch to using `@org.springframework.boot.autoconfigure.condition.ConditionalOnBooleanProperty` annotation?

robertonav20:
hi @tadayosi and @grgrzybek ,
i'm agree with your comments, i will work on it as soon as possible

i didn't know this `ConditionalOnBooleanProperty` annotation, nice advice!

grgrzybek:
Thanks - any chance to have a PR with a single commit? (no `merge branch` commits please).

robertonav20:
Hi @tadayosi 
i'm trying to reach your goal on quarkus and war 

Quarkus
i successfully disabled all apis less hawtio/index.html
{DFAFF63E-674E-44AF-9185-255FBEC31567}

War
I don't know where/how i should apply this feature

Could you give me some advice or suggestion?

Thanks in advance

grgrzybek:
@robertonav20 - please don't bother with _disabling Hawtio in WAR_. Disabled WAR is just no WAR ;)

robertonav20:
I'm honestly confused about your last comment 🥴

grgrzybek:
@robertonav20 simply - I agree that `hawtio.enabled` property in "Hawtio SpringBoot", together with `@ConditionalOnBooleanProperty` is fine and in accordance to Spring Boot ideas.

I don't think we need an equivalent of this property (and behavior) for WAR deployment.

tadayosi:
@robertonav20 @grgrzybek Sorry for the confusion. I now agree with you that the flag makes sense only for Spring Boot. So please go back to the initial PR and we are ready to merge it. Please ignore war and Quarkus. Thank you.

PR #4162, opened 2025-11-27
[#HWT4155] Configuration - added property to manage activation

robertonav20:
Hi
this is my PR to manage the scenario of https://github.com/hawtio/hawtio/issues/4155.

i added a new property hawtio.enabled to manage hawtio activation default is true

Origin PR https://github.com/hawtio/hawtio/pull/4157

### DIFF
--- a/deploy/hawtio-default/src/main/java/io/hawt/HawtioDefaultContextListener.java
+++ b/deploy/hawtio-default/src/main/java/io/hawt/HawtioDefaultContextListener.java
@@ -2,7 +2,6 @@ package io.hawt;
 
 import jakarta.servlet.ServletContextEvent;
 
-import io.hawt.HawtioContextListener;
 import io.hawt.jvm.local.JVMList;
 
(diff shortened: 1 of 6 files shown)


grgrzybek:
I think `hawtio.enabled` for Spring Boot deployments is ok.
I don't know exactly if we need similar flag for Quarkus...

I think you needed this flag in Spring Boot to be able to change the behavior after building Hawtio Spring Boot (prod, dev environments).

But for Quarkus, you have this flag at build time, so you need two builds anyway - with and without this flag... So I don't think we need it.
WDYT @tadayosi @jamesnetherton ?

robertonav20:
Hi @grgrzybek @tadayosi 
I will come back to previous commit, remove quakrua changes and repush with force if you are agree

tadayosi:
I agree with going only for spring boot.

robertonav20:
Hi @grgrzybek @tadayosi 
PR updated as requested

grgrzybek:
Let's merge. It's good now.

PR #4338, opened 2026-02-27
E2E diagnostics tests

jsolovjo:
In this PR:

- New JFR tests implemented by @joshiraezcode;
- Due to some limitations inside containers running on GH Actions, the JFR is disabled, so it's better to skip the tests;
- The tests are run only in JBang workflow and when triggered on a local machine;
- We had to add a new env var in the e2e workflow file, so the current PR's checks will fail;

The successful run - https://github.com/jsolovjo/hawtio/actions/runs/22485164490

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -94,6 +94,7 @@ jobs:
             extra_args='-Dio.hawt.test.use.keycloak=true'
           fi
           docker run --rm \
+          -e GITHUB_ACTIONS=true \
           -v /var/run/docker.sock:/var/run/docker.sock \
           -v $PWD/$REPORT_DIR:/workspace/tests/hawtio-test-suite/target \
           -v $PWD/$REPORT_DIR/build:/workspace/tests/hawtio-test-suite/build/ \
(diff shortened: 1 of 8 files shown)


joshiraezcode:
Juri, let me do a PR against your branch because I'm renaming the Diagnostics plugin and tests will fail

jsolovjo:
Rebased, checked and tested in my forked project - https://github.com/jsolovjo/hawtio/actions/runs/23340439786

jsolovjo:
Most likely, the checks will fail as I set an env var here - https://github.com/hawtio/hawtio/pull/4338/changes#diff-a7287cabb55032a1b0277a08d47a1c867352ff9066be89908f0b119f5aa71463R97 which helps to skip the tests if they are run in github actions https://github.com/hawtio/hawtio/pull/4338/changes#diff-084fb99f2585c3edeebb1e8be02f677731b8e4f66cf12b696eee8be15c5da23bR55

As we touched the e2e workflow file, it won't be applied and won't be working until it's merged.

tadayosi:
let's merge it then 🤞

PR #4388, opened 2026-03-22
task: do not execute SessionExpiryFilter for every call

GChuf:
I'm submitting this for v4.x only for now.

I was profiling some code and I noticed hawtio's SessionExpiryFilter.doFilter uses a very big amount of CPU compared to other methods. I was testing this inside apache artemis web console. While the broker was idle, and the webpage refreshes were the only thing using CPU, SessionExpiryFilter used 50% of cpu time.

image

I realized this filter does some "work" for every call - even for static files.
(only) after that, it filters out _IGNORED_PATHS_.

I don't think executing this for every file/call is the correct way to approach this.

I wasn't sure know which URL patterns should implement SessionExpiryFilter - but for v4, at least, I came to the conclusion that "refresh" calls are the only (and a very good) indicator if a user is "active" on the webpage.

Tested latest v4 only, through artemis. Session expiry behaves as expected. Results:

image

Let me know what you think ...

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/auth/SessionExpiryFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/auth/SessionExpiryFilter.java
@@ -39,8 +39,6 @@ public class SessionExpiryFilter implements Filter {
 
     public static final String ATTRIBUTE_LAST_ACCESS = "LastAccess";
 
-    private static final List<String> IGNORED_PATHS = List.of("jolokia", "proxy");
-
     private AuthenticationConfiguration authConfiguration;
     private int pathIndex;
 
@@ -128,11 +126,8 @@ public class SessionExpiryFilter implements Filter {
         }
 
         LOG.debug("SubContext: {}", subContext);
-        if (IGNORED_PATHS.contains(subContext) && session.getAttribute(ATTRIBUTE_LAST_ACCESS) != null) {
-            LOG.debug("Not updating LastAccess");
-        } else {
-            updateLastAccess(session, now);
-        }
+
+        updateLastAccess(session, now);
 
         chain.doFilter(request, response);
     }
--- a/platforms/springboot/src/main/java/io/hawt/springboot/HawtioManagementConfiguration.java
+++ b/platforms/springboot/src/main/java/io/hawt/springboot/HawtioManagementConfiguration.java
@@ -183,7 +183,7 @@ public class HawtioManagementConfiguration {
     public FilterRegistrationBean<SessionExpiryFilter> sessionExpiryFilter() {
         final FilterRegistrationBean<SessionExpiryFilter> filter = new FilterRegistrationBean<>();
         filter.setFilter(new SessionExpiryFilter());
-        filter.addUrlPatterns(hawtioPath + "/*");
+        filter.addUrlPatterns(hawtioPath + "/refresh");
         return filter;
     }
 


tadayosi (commented):
I'm still not convinced that this is really the right solution for the observed issue. I checked the behaviour of Hawtio (not Artemis) but the static files are requested only once at the initial page load and cache are hit afterwards. In most case, the user activities are represented as requests to `/jolokia` (which might be another issue considering it's included in the ignored path... but it's a separate issue) and they aren't followed by access to static resources.

tadayosi:
@GChuf As commented above, at this moment I'm not sure if it's really an issue with Hawtio. Isn't it an issue specific to Artemis Console? By any chance, doesn't any custom HTTP header added by Artemis cause the static resources to be requested every time? Could you check if the same issue can be really observed with a vanilla Hawtio instance?

@grgrzybek Please review this and let us know what you think too.

grgrzybek:
Thanks for notification - I'll check it today.
I remember this filter was very important for the purpose of authentication... Let me review.

tadayosi:
@grgrzybek thanks. I suspect that applying the filter to `/refresh` even has been a mistake.

GChuf:
@tadayosi thanks for your response.
The issue is not that the static files are reloaded every time, it's that the expensive SessionExpiryFilter is being applied for all static files as well.
I just confirmed this issue is present in vanilla hawtio v4 as well. This is the result of me refreshing (not hard refreshing) the "connect" screen a couple of times, after logging in:

image

After logging in to the broker and clicking around the menu, the situation is "a bit better" - probably because the filter isn't applied on all the jolokia calls?
image




Since jolokia calls are being executed automatically, applying the filter to /jolokia calls could result in sessions never expiring, if I understand this correctly. While "refresh" calls only happen after user actions on the GUI. At least that was my experience.

grgrzybek:
There are two roles played by SessionExpiryFilter:
 - to return `ok` if the session is valid
 - to actually invalidate the session if there's _any_ (user = /refresh or automatic = /jolokia) activity after the session (as calculated manually - not due to normal Servlet API usage)

Mind that this screen is misleading:
image

you see 96% for SessionExpiryFilter because part of _each_ filter code is to call the chain - and it's the chain (which includes consecutive filters as well as the target servlet) that takes almost all the time.

SessionExpiryFilter is cheap - just a map access and `(now - lastAccess) / 1000` calculation...

GChuf:
@grgrzybek you're right - I was looking at those graphs too fast today, and yesterday I assumed that SessionExpiryFilter is the last call.
While trying to reproduce - I remembered that I only saw those huge differences in numbers yesterday because I was hard-refreshing the page - forcing the static content to be redownloaded.
The above tests are results of just clicking around the webpages.

While trying to reproduce the results from yesterday by doing hard refreshes (huge difference in cpu usage for FilterChain), I was able to reproduce the result, but not reliably. So I'm not sure what's going on.

That said, even if this idea does not improve performance or speed, I still stand by my initial idea. The session still gets expired on no activity, or prolonged after "refresh " is called.

tadayosi:
> to actually invalidate the session if there's any (user = /refresh or automatic = /jolokia) activity after the session (as calculated manually - not due to normal Servlet API usage)

Ah ok, that was my misunderstanding then. It makes sense to handle /refresh requests. /jolokia has been in the IGNORED_PATHS, so it only touches the initial last access timestamp but never updates it afterwards.

@GChuf 

> The issue is not that the static files are reloaded every time, it's that the expensive SessionExpiryFilter is being applied for all static files as well.

I see it the other way round. If the filter is only applied at the beginning but never triggered afterwards thanks to caching, as Grzegorz says, it's just a matter of doing some calculations so it shouldn't be too much load as a whole. If the filter is applied for every request on all static files even though you haven't done hard reload on the browser, that should be a problem. That's what I'm saying.

GChuf:
@tadayosi 
> I see it the other way round. If the filter is only applied at the beginning but never triggered afterwards thanks to caching, as Grzegorz says, it's just a matter of doing some calculations so it shouldn't be too much load as a whole. If the filter is applied for every request on all static files even though you haven't done hard reload on the browser, that should be a problem. That's what I'm saying.

I understand what you're saying. My initial thoughts were that the filter is expensive, and so any extra work by the filter should be avoided, even if only on first load.

grgrzybek:
I'd simply say - this filter is really not expensive and there's really not THAT many static resources served by Hawtio. When full app is loaded, it's only `/jolokia` endpoint.

Possibly we should not "update last access" when getting static resources, but this is cheap:
```java
session.setAttribute(ATTRIBUTE_LAST_ACCESS, now)
```

You're intuition was mostly ok when changing:
```java
filter.addUrlPatterns(hawtioPath + "/*");
```
to
```java
filter.addUrlPatterns(hawtioPath + "/refresh");
```

but hitting CTRL-R by user is _also_ a sign that the session should be refreshed - that's why I think it's not necessary to change anything here.

PR #4440, opened 2026-04-13
feat(system): make CSP Filter extendable by users

### LINKED ISSUE #4437: Make CSP Filter extendable by users
Currently, by default Hawtio sends the following `Content-Security-Policy` http header for security reasons:

> Content-Security-Policy: default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src 'self' data:; connect-src 'self'; frame-src 'self'; manifest-src 'self'; media-src 'self'; object-src 'self'; worker-src 'self' blob:; frame-ancestors 'none'; script-src-elem 'self'; style-src-elem 'self' 'unsafe-inline'; form-action 'self'

From a security perspective, this restrictive CSP is excellent; however, the restrictions on `connect-src` and `script-src` in particular hinder support for more advanced external plugins that can extend Hawtio's functionality further.

Provided that a plugin’s functionality is clearly explained and the user has a full understanding of which external resources the plugin may access and deems it safe, that plugin should be runnable.

For these reasons, the CSP Filter should be made extensible by the user. Specifically, it should be possible to add trusted external resources to `connect-src` and `script-src` as required via system properties.

Related issue: #3225

tadayosi:
Fix #4437

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/web/filters/ContentSecurityPolicyFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/filters/ContentSecurityPolicyFilter.java
@@ -6,9 +6,9 @@ import java.io.InputStreamReader;
 import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
-
 import jakarta.servlet.FilterConfig;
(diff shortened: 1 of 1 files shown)


grgrzybek (approved):
👍

PR #4457, opened 2026-04-28
feat(auth): Use new Subject API under JDK 18+ (Fixes #4456)

### LINKED ISSUE #4456: Hawtio's RBAC Authentication still uses deprecated Subject.doAs() and Subject.getSubject() calls in AuthenticationFilter and RBACMBeanInvoker
### Hawtio's RBAC Authentication still uses deprecated Subject.doAs() and Subject.getSubject() calls in AuthenticationFilter and RBACMBeanInvoker

**Subject.doAs(...) + Subject.getSubject(...)** may be replaced by **Subject.callAs(...) + Subject.current()** as below.

**io.hawt.system.RBACMBeanInvoker**

```
    public boolean canInvoke(ObjectName objectName, String operation) {
        if (this.securityMBean == null) {
            // JMXSecurity MBean is not found, thus RBAC is disabled
            return true;
        }

        Subject subject = Subject.current();
        try {
            if (subject != null) {
                String username = AuthHelpers.getUsername(subject);
                return canInvokeCache.get(new CanInvokeKey(username, objectName, operation));
            } else {
                // For now, let's bypass the caching and directly invoke the security MBean if
                // subject is not available (which could happen on some platforms other than Karaf)
                LOG.debug("Subject not available, directly invoking canInvoke(): {}, {}", objectName, operation);
                return doCanInvoke(objectName, operation);
            }
        } catch (Exception e) {
            logMBeanError(e);
            return false;
        }
    }

```

**io.hawt.web.auth.AuthenticationFilter**

```
    private static void executeAs(final ServletRequest request, final ServletResponse response, final FilterChain chain, Subject subject) {
        try {
            Subject.callAs(subject, () -> {
                chain.doFilter(request, response);
                return null;
            });
        } catch (Exception e) {
            LOG.info("Failed to handle {} due to:", ((HttpServletRequest) request).getRequestURI(), e);
        }
    }

```

----
Related issues:
* jolokia/jolokia#1002

grgrzybek:


joshiraezcode:
@grgrzybek Put the Fixes #4456 inside the PR description so it autolinks. I think PR/commit messages are ignored, although it's useful to reference it there as well

grgrzybek:
> @grgrzybek Put the Fixes #4456 inside the PR description so it autolinks. I think PR/commit messages are ignored, although it's useful to reference it there as well

hmm - I used this in my commit message ;)

joshiraezcode (approved):
After checking the new Subject API I think it's correct. Fenomenal work.

joshiraezcode:
@grgrzybek Yes I know, I was just telling that the site ignores autolinking the PR to the issue if it's not on the PR description even though it correctly links on the whole site (you can check the PR that is a link for the issue).

I have set the referred task manually

tadayosi (approved):
elegant!

grgrzybek:
it was about time to check what is `MethodHandle` ;)

PR #4482, opened 2026-05-05
[E2E] Add new operator namespace watching tests.

jsolovjo:
In this PR:
- Added new Hawtio Online tests which cover AllNamespaces, OwnNamespace, SingleNamespace installation modes;
- Stabilized a few flakiness in the Operator tests;
- Enabled the RBAC test
- - Fixed ACL yaml file as it wasn't correct in the test suite
- - Some UI elements were changed, so it required a small refactoring and fixes in the assertions/checks
- Removed a custom logo test as previously an external source was used and now it's blocked by Content Source Policy, therefore, we will need to refactor the test later and for now we temporarily disabled it to avoid test run failures;

Checked it in our environment and all tests pass - `Tests run: 140, Failures: 0, Errors: 0, Skipped: 6`

### DIFF
--- a/tests/hawtio-test-suite/src/test/resources/io/hawt/tests/openshift/acl.yaml
+++ b/tests/hawtio-test-suite/src/test/resources/io/hawt/tests/openshift/acl.yaml
@@ -179,7 +179,7 @@ org.apache.camel.context:
   getLoad15: admin, viewer
   getCamelVersion: admin, viewer
   getApplicationContextClassName: admin, viewer
-#  getTotalRoutes: admin, viewer
+  #  getTotalRoutes: admin, viewer
   getStartedRoutes: admin, viewer
   isMessageHistory: admin, viewer
   getTimeUnit: admin, viewer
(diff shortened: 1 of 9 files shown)


tests/hawtio-test-suite/src/test/resources/io/hawt/tests/openshift/acl.yaml
```diff
@@ -279,6 +279,11 @@ org.apache.camel.tracer:
   clear: admin
   resetTraceCounter: admin
 
+# Catch-all for org.apache.camel domain (evaluated LAST after all specific rules above)
+# Using explicit roles instead of /.*/ regex for better security
+org.apache.camel:
+  /.*/: admin, viewer
```
tadayosi: Does the order change the behaviour?  Or is it only for human readability?

tadayosi (approved):

tests/hawtio-test-suite/src/test/resources/io/hawt/tests/openshift/acl.yaml
jsolovjo: That's the interesting question. 
Perhaps, it was wrong from the beginning as this
```
org.apache.camel:
  /.*/: admin, viewer
```
was applied first and then the `stop` operation was still enabled for both - admin and viewer users. 
The following `stop: []` was simply ignored, and when I changed the order - the rbac started working as expected and the `stop` operation became disabled.

jsolovjo: or you mean that `/.*/: /.*/` was replaced by `/.*/: admin, viewer` ? In this case, it's just for for human readability and our test cluster has only admin and viewer users

tadayosi (approved):

PR #4489, opened 2026-05-11
[E2E] Update CRD files for testing purpose.

jsolovjo:
Updates for CRD files which are used in e2e testing.


tadayosi (approved):
Don't we need it upstream (main) as well?

jsolovjo:
> Don't we need it upstream (main) as well?

yes, right
I will create another PR too

PR #4556, opened 2026-06-25
[E2E] Update the checkout workflow files.

jsolovjo:
In this PR:
- A security advisory suggests to use pull_request
- As GitHub strips out all secrets and all tokens are strictly read-only, the steps using the tokens will fail immediately
  - As a temporal solution is to disable them, so no summary comments in the PR for a while
  - And then refactor the workflow files to use a two-workflow approach which is a common practice nowadays
- The updated workflow files have been tested here (passed):
  - e2e: https://github.com/jsolovjo/hawtio/actions/runs/28160696287
  - jbang: https://github.com/jsolovjo/hawtio/actions/runs/28160696203

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -1,7 +1,7 @@
 name: E2E Test
 
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - main
       - 4.x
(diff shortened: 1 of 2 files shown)


tadayosi (approved):
thanks!

PR #4560, opened 2026-06-29
feat: Separate unauthorized/forbidden statuses

GChuf:
I've added "forbidden" and "authorized" statuses - and changed the NO_CREDENTIALS case to unauthorized instead of forbidden - according to the [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401):
`A 401 Unauthorized is similar to the 403 Forbidden response, except that a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action.`

The main goal was to have a bit more information on failed logins to the console - is the password wrong, or just the user roles?
Linked to https://github.com/hawtio/hawtio-react/pull/2152

### DIFF
--- a/hawtio-system/src/main/java/io/hawt/system/AuthenticateResult.java
+++ b/hawtio-system/src/main/java/io/hawt/system/AuthenticateResult.java
@@ -8,6 +8,7 @@ public class AuthenticateResult {
         AUTHORIZED,
         NOT_AUTHORIZED,
         NO_CREDENTIALS,
+        FORBIDDEN,
         THROTTLED
     }
 
@@ -17,6 +18,10 @@ public class AuthenticateResult {
 
     public static AuthenticateResult notAuthorized() {
         return new AuthenticateResult(Type.NOT_AUTHORIZED);
+    }
+
+    public static AuthenticateResult forbidden() {
+        return new AuthenticateResult(Type.FORBIDDEN);
     }
 
     public static AuthenticateResult noCredentials() {
--- a/hawtio-system/src/main/java/io/hawt/system/Authenticator.java
+++ b/hawtio-system/src/main/java/io/hawt/system/Authenticator.java
@@ -199,6 +199,7 @@ public class Authenticator {
 
     public AuthenticateResult authenticate(Consumer<Subject> callback) {
         if (hasNoCredentials()) {
+            LOG.debug("Authentication failed: No credentials provided");
             return AuthenticateResult.noCredentials();
         }
 
@@ -213,47 +214,48 @@ public class Authenticator {
             return AuthenticateResult.throttled(attempt.retryAfter());
         }
 
-        Subject subject = doAuthenticate();
-        if (subject == null) {
-            throttler.ifPresent(t -> t.increase(username));
-            return AuthenticateResult.notAuthorized();
-        }
-        throttler.ifPresent(t -> t.reset(username));
-
-        if (callback != null) {
-            try {
-                callback.accept(subject);
-            } catch (Exception e) {
-                LOG.warn("Failed to execute privileged action:", e);
-            }
-        }
-
-        return AuthenticateResult.authorized();
-    }
-
-    protected Subject doAuthenticate() {
         String realm = authConfiguration.getRealm();
         List<String> roles = authConfiguration.getRoles();
         List<Class<Principal>> rolePrincipalClasses = authConfiguration.getRolePrincipalClasses();
         Configuration configuration = authConfiguration.getConfiguration();
 
+        Subject subject = new Subject();
+
         try {
             LOG.debug("doAuthenticate[realm={}, roles={}, rolePrincipalClasses={}, configuration={}, username={}, password={}]",
                 realm, String.join(", ", roles), rolePrincipalClasses, configuration, username, "******");
 
-            Subject subject = new Subject();
             login(subject, realm, configuration);
-            if (checkRoles(subject, roles, rolePrincipalClasses)) {
-                return subject;
+            if (!checkRoles(subject, roles, rolePrincipalClasses)) {
+                LOG.warn("Login failed due to role mismatch for user: {}", username);
+                throttler.ifPresent(t -> t.increase(username));
+                return AuthenticateResult.forbidden();
             }
         } catch (AccountException e) {
             LOG.warn("Account failure", e);
+            throttler.ifPresent(t -> t.increase(username));
+            return AuthenticateResult.notAuthorized();
         } catch (LoginException e) {
             LOG.warn("Login failed due to: {}", e.getMessage());
+            throttler.ifPresent(t -> t.increase(username));
             LOG.debug("Failed stacktrace:", e);
+            return AuthenticateResult.notAuthorized();
+        } catch (Exception e) {
+            LOG.error("Unexpected error during authentication: {}", e);
+            throttler.ifPresent(t -> t.increase(username));
+            return AuthenticateResult.notAuthorized();
         }
 
-        return null;
+        if (callback != null) {
+            try {
+                callback.accept(subject);
+            } catch (Exception ex) {
+                LOG.warn("Failed to execute privileged action:", ex);
+            }
+        }
+
+        throttler.ifPresent(t -> t.reset(username));
+        return AuthenticateResult.authorized();
     }
 
     protected void login(Subject subject, String realm, Configuration configuration) throws LoginException {
--- a/hawtio-system/src/main/java/io/hawt/system/Authenticator.java
+++ b/hawtio-system/src/main/java/io/hawt/system/Authenticator.java
@@ -15,7 +15,6 @@ import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.NameCallback;
 import javax.security.auth.callback.PasswordCallback;
-import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.auth.login.AccountException;
 import javax.security.auth.login.Configuration;
 import javax.security.auth.login.LoginContext;
@@ -214,48 +213,56 @@ public class Authenticator {
             return AuthenticateResult.throttled(attempt.retryAfter());
         }
 
+        Subject subject = new Subject();
+        AuthenticateResult authResult = doAuthenticate(subject);
+
+        if (authResult.getType() == AuthenticateResult.Type.AUTHORIZED) {
+            throttler.ifPresent(t -> t.reset(username));
+            if (callback != null) {
+                try {
+                    callback.accept(subject);
+                } catch (Exception ex) {
+                    LOG.warn("Failed to execute privileged action:", ex);
+                }
+            }
+        } else {
+            throttler.ifPresent(t -> t.increase(username));
+        }
+
+        return authResult;
+    }
+
+
+    protected AuthenticateResult doAuthenticate(Subject subject) {
         String realm = authConfiguration.getRealm();
         List<String> roles = authConfiguration.getRoles();
         List<Class<Principal>> rolePrincipalClasses = authConfiguration.getRolePrincipalClasses();
         Configuration configuration = authConfiguration.getConfiguration();
-
-        Subject subject = new Subject();
 
         try {
             LOG.debug("doAuthenticate[realm={}, roles={}, rolePrincipalClasses={}, configuration={}, username={}, password={}]",
                 realm, String.join(", ", roles), rolePrincipalClasses, configuration, username, "******");
 
             login(subject, realm, configuration);
-            if (!checkRoles(subject, roles, rolePrincipalClasses)) {
-                LOG.warn("Login failed due to role mismatch for user: {}", username);
-                throttler.ifPresent(t -> t.increase(username));
+            if (checkRoles(subject, roles, rolePrincipalClasses)) {
+                return AuthenticateResult.authorized();
+            } else {
+                LOG.debug("Login failed due to role mismatch for user: {}", username);
                 return AuthenticateResult.forbidden();
             }
         } catch (AccountException e) {
             LOG.warn("Account failure", e);
-            throttler.ifPresent(t -> t.increase(username));
             return AuthenticateResult.notAuthorized();
         } catch (LoginException e) {
             LOG.warn("Login failed due to: {}", e.getMessage());
-            throttler.ifPresent(t -> t.increase(username));
             LOG.debug("Failed stacktrace:", e);
             return AuthenticateResult.notAuthorized();
         } catch (Exception e) {
             LOG.error("Unexpected error during authentication: {}", e);
-            throttler.ifPresent(t -> t.increase(username));
+            LOG.debug("Failed stacktrace:", e);
             return AuthenticateResult.notAuthorized();
         }
 
-        if (callback != null) {
-            try {
-                callback.accept(subject);
-            } catch (Exception ex) {
-                LOG.warn("Failed to execute privileged action:", ex);
-            }
-        }
-
-        throttler.ifPresent(t -> t.reset(username));
-        return AuthenticateResult.authorized();
     }
 
     protected void login(Subject subject, String realm, Configuration configuration) throws LoginException {
--- a/hawtio-system/src/main/java/io/hawt/web/ServletHelpers.java
+++ b/hawtio-system/src/main/java/io/hawt/web/ServletHelpers.java
@@ -36,7 +36,7 @@ public class ServletHelpers {
     private static final JolokiaSerializer SERIALIZER = new JolokiaSerializer();
 
     public static void doForbidden(HttpServletResponse response) {
-        doForbidden(response, ForbiddenReason.NONE);
+        doForbidden(response, ForbiddenReason.FORBIDDEN);
     }
 
     public static void doForbidden(HttpServletResponse response, ForbiddenReason reason) {
@@ -47,6 +47,21 @@ public class ServletHelpers {
             response.flushBuffer();
         } catch (IOException ioe) {
             LOG.debug("Failed to send forbidden response: {}", ioe.toString());
+        }
+    }
+
+    public static void doUnauthorized(HttpServletResponse response) {
+        doUnauthorized(response, ForbiddenReason.NOT_AUTHORIZED);
+    }
+
+    public static void doUnauthorized(HttpServletResponse response, ForbiddenReason reason) {
+        try {
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            response.setHeader(HEADER_HAWTIO_FORBIDDEN_REASON, reason.name());
+            response.setContentLength(0);
+            response.flushBuffer();
+        } catch (IOException ioe) {
+            LOG.debug("Failed to send unauthorized response: {}", ioe.toString());
         }
     }
 
--- a/hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationFilter.java
+++ b/hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationFilter.java
@@ -144,16 +144,19 @@ public class AuthenticationFilter implements Filter {
             // request was already executed using the authenticated subject in executeAs(), nothing more to do
             break;
         case NOT_AUTHORIZED:
-            ServletHelpers.doForbidden(httpResponse);
+            ServletHelpers.doUnauthorized(httpResponse);
             break;
         case NO_CREDENTIALS:
             if (authConfiguration.isNoCredentials401()) {
                 // return auth prompt 401
                 ServletHelpers.doAuthPrompt(httpResponse, authConfiguration.getRealm());
             } else {
-                // return forbidden 403 so the browser login does not popup
-                ServletHelpers.doForbidden(httpResponse);
+                // return unauthorized 401 so the browser login does not popup
+                ServletHelpers.doUnauthorized(httpResponse);
             }
+            break;
+        case FORBIDDEN:
+            ServletHelpers.doForbidden(httpResponse);
             break;
         case THROTTLED:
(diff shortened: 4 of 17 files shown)


GChuf:
Ready for proper review.
doAuthenticate() was inlined, a couple of console log levels were changed. Things can be changed, it's up to you/your preferences.

The behavior after these changes is:

| Scenario | HTTP Status | `Hawtio-Forbidden-Reason` | Frontend info | Console log |
|---|---|---|---|---|
| First visit | 403 | `FORBIDDEN` | N/A | N/A |
| Wrong password | 401 | `NOT_AUTHORIZED` | Invalid login credentials | `Login failed due to: Password does not match for user: ...` |
| User does not exist | 401 | `NOT_AUTHORIZED` | Invalid login credentials | `Login failed due to: User does not exist: ...` |
| Empty username / password / both | 401 | `NOT_AUTHORIZED` | N/A | `Authentication failed: No credentials provided` |
| Wrong roles | 403 | `FORBIDDEN` | Access forbidden | `Login failed due to role mismatch for user: ...` |
| Throttled | 429 | N/A | Login attempt blocked. Retry in ... | `Authentication throttled: ...` |

grgrzybek:
> I've added "forbidden" and "authorized" statuses - and changed the NO_CREDENTIALS case to unauthorized instead of forbidden - according to the [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401):
> `A 401 Unauthorized is similar to the 403 Forbidden response, except that a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action.`

Indeed the docs are a bit surprising, suggesting, that 403 is about failed _authorization_.

But additionally MDN says:

> Clients that receive a 403 response should expect that repeating the request without modification will fail with the same error

I always used this rule:
 - 401 to tell the caller that you should authenticate
 - 403 to tell the caller that authentication (or authorization) failed with the passed credentials - both if the credentials are wrong or if the credentials are for a user that doesn't have sufficient permissions.

But anyway - let me review.

hawtio-system/src/main/java/io/hawt/web/auth/LoginServlet.java
```diff
@@ -80,7 +81,12 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
                 break;
             case NOT_AUTHORIZED:
+                ServletHelpers.doUnauthorized(response);
+                break;
             case NO_CREDENTIALS:
+                ServletHelpers.doUnauthorized(response);
+                break;
+            case FORBIDDEN:
                 ServletHelpers.doForbidden(response);
                 break;
```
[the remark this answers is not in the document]
GChuf: ServletHelpers.doForbidden(response) calls ServletHelpers.doForbidden(response, ForbiddenReason.FORBIDDEN), similar to every other case.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/servlets/HawtioQuakusLoginServlet.java
```diff
@@ -43,12 +44,17 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
         AuthenticateResult result = authenticator.authenticate(authConfiguration, username, password);
         switch (result.getType()) {
         case AUTHORIZED:
-            LOG.info("Logging in user: {}", username);
             AuthSessionHelpers.setup(request.getSession(true), new Subject(), username, timeout);
+            LOG.info("Logging in user: {}", username);
+            break;
+        case FORBIDDEN:
+            ServletHelpers.doForbidden(response);
             break;
```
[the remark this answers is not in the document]
GChuf: ServletHelpers.doForbidden(response) calls ServletHelpers.doForbidden(response, ForbiddenReason.FORBIDDEN), similar to every other case.

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
```diff
@@ -209,51 +210,52 @@ public AuthenticateResult authenticate(Consumer<Subject> callback) {
...
         } catch (LoginException e) {
             LOG.warn("Login failed due to: {}", e.getMessage());
+            throttler.ifPresent(t -> t.increase(username));
             LOG.debug("Failed stacktrace:", e);
+            return AuthenticateResult.notAuthorized();
+        } catch (Exception e) {
+            LOG.error("Unexpected error during authentication: {}", e);
+            throttler.ifPresent(t -> t.increase(username));
+            return AuthenticateResult.notAuthorized();
         }
```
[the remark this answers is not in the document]
GChuf: @grgrzybek @tadayosi  this one is up to you - I think it makes sense.

hawtio-system/src/main/java/io/hawt/system/AuthenticateResult.java
```diff
@@ -8,7 +8,8 @@ public enum Type {
         AUTHORIZED,
         NOT_AUTHORIZED,
         NO_CREDENTIALS,
-        THROTTLED
+        THROTTLED,
+        FORBIDDEN
     }
```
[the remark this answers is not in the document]
GChuf: Addressed this in 4f6e97d0c9a9a84262114b80d5e5f0ba10bd3538

hawtio-system/src/main/java/io/hawt/web/ServletHelpers.java
```diff
@@ -36,7 +36,7 @@ public class ServletHelpers {
     private static final JolokiaSerializer SERIALIZER = new JolokiaSerializer();
 
     public static void doForbidden(HttpServletResponse response) {
-        doForbidden(response, ForbiddenReason.NONE);
+        doForbidden(response, ForbiddenReason.FORBIDDEN);
     }
```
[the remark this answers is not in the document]
GChuf: Not sure about this one, but since NONE was now only used in the tests, I removed it in ba5d0818e8cf5f083518b7425a9309717d878cd4.

hawtio-system/src/main/java/io/hawt/system/AuthenticateResult.java
```diff
@@ -8,7 +8,8 @@ public enum Type {
         AUTHORIZED,
         NOT_AUTHORIZED,
         NO_CREDENTIALS,
-        THROTTLED
+        THROTTLED,
+        FORBIDDEN
```
tadayosi: `FORBIDDEN` can be put higher than `THROTTLED` as `THROTTLED` is different from the other statuses.

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
```diff
@@ -199,6 +199,7 @@ public static void logout(AuthenticationConfiguration authConfiguration, Subject
 
     public AuthenticateResult authenticate(Consumer<Subject> callback) {
         if (hasNoCredentials()) {
+            LOG.warn("Authentication failed: No credentials provided");
```
tadayosi: Why do we want this warning log?
I'd prefer to be more conservative on increasing the amount of logs unless there's a reason.

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
```diff
@@ -209,51 +210,52 @@ public AuthenticateResult authenticate(Consumer<Subject> callback) {
             .filter(AuthenticationThrottler.Attempt::isBlocked)
             .orElse(null);
         if (attempt != null) {
-            LOG.debug("Authentication throttled: {}", attempt);
+            LOG.warn("Authentication throttled: {}", attempt);
```
tadayosi: The same applies here. The throttling logs are intentionally set to debug because they can overwhelm the log file.

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
```diff
@@ -209,51 +210,52 @@ public AuthenticateResult authenticate(Consumer<Subject> callback) {
...
-                callback.accept(subject);
-            } catch (Exception e) {
-                LOG.warn("Failed to execute privileged action:", e);
-            }
-        }
-
-        return AuthenticateResult.authorized();
-    }
-
-    protected Subject doAuthenticate() {
```
tadayosi, line 234: Why is this removed?  I don't like this merger of a separate protected method.

Can't we do this while keeping `doAuthenticate()`?

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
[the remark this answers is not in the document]
tadayosi: no need for addressing it

hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationFilter.java
```diff
@@ -144,17 +144,20 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
             break;
         case NO_CREDENTIALS:
             if (authConfiguration.isNoCredentials401()) {
                 // return auth prompt 401
                 ServletHelpers.doAuthPrompt(httpResponse, authConfiguration.getRealm());
             } else {
-                // return forbidden 403 so the browser login does not popup
-                ServletHelpers.doForbidden(httpResponse);
+                // return unauthorized 401 so the browser login does not popup
+                ServletHelpers.doUnauthorized(httpResponse);
```
tadayosi: This doesn't seem correct, as now both if/else clauses return `401`. 401 was used to trigger auth prompt according to the comment.

@grgrzybek What do you think about this?

hawtio-system/src/main/java/io/hawt/web/ServletHelpers.java
```diff
@@ -36,7 +36,7 @@ public class ServletHelpers {
     private static final JolokiaSerializer SERIALIZER = new JolokiaSerializer();
 
     public static void doForbidden(HttpServletResponse response) {
-        doForbidden(response, ForbiddenReason.NONE);
+        doForbidden(response, ForbiddenReason.FORBIDDEN);
```
tadayosi, line 39: I don't remember the purpose of `NONE` but I guess it serves as a default reason. We probably need to remember if this legacy part has a real meaning in the code.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/auth/HawtioQuarkusAuthenticator.java
```diff
@@ -44,7 +44,7 @@ public AuthenticateResult authenticate(AuthenticationConfiguration authConfigura
             .filter(AuthenticationThrottler.Attempt::isBlocked)
             .orElse(null);
         if (attempt != null) {
-            LOG.debug("Authentication throttled: {}", attempt);
+            LOG.warn("Authentication throttled: {}", attempt);
```
tadayosi: do we need the change?

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/auth/HawtioQuarkusAuthenticator.java
```diff
@@ -56,7 +56,8 @@ public AuthenticateResult authenticate(AuthenticationConfiguration authConfigura
             List<String> roleConfig = authConfiguration.getRoles();
             // Verify the allowed roles matches with those specified in Quarkus security config
             if (!verifyRole(identity, roleConfig)) {
-                return AuthenticateResult.notAuthorized();
+                LOG.warn("Login failed due to role mismatch for user: {}", username);
```
tadayosi, line 59: the same applies here. do we need logging?

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/filters/HawtioQuarkusAuthenticationFilter.java
```diff
@@ -133,17 +133,20 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
             break;
         case NO_CREDENTIALS:
             if (authConfiguration.isNoCredentials401()) {
                 // return auth prompt 401
                 ServletHelpers.doAuthPrompt(httpResponse, authConfiguration.getRealm());
             } else {
-                // return forbidden 403 so the browser login does not popup
-                ServletHelpers.doForbidden(httpResponse);
+                // return unauthorized 401 so the browser login does not popup
+                ServletHelpers.doUnauthorized(httpResponse);
```
tadayosi: the same applies here

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/servlets/HawtioQuarkusLoginServlet.java
```diff
@@ -43,12 +44,17 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
         AuthenticateResult result = authenticator.authenticate(authConfiguration, username, password);
         switch (result.getType()) {
         case AUTHORIZED:
-            LOG.info("Logging in user: {}", username);
             AuthSessionHelpers.setup(request.getSession(true), new Subject(), username, timeout);
+            LOG.info("Logging in user: {}", username);
```
tadayosi, line 47: why the change?

tadayosi (changes requested):
Sorry for being late. Thanks for your contribution.

There are a few places that I want you to review.

hawtio-system/src/main/java/io/hawt/system/AuthenticateResult.java
GChuf: resolved

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
GChuf: I had some issues here which were solved by inlining doAuthenticate, I cannot remember exactly what.
I will rewrite this part so that doAuthenticate stays.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/servlets/HawtioQuarkusLoginServlet.java
GChuf: Reverted this one.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/auth/HawtioQuarkusAuthenticator.java
GChuf: The idea was to give more info to the administrator - i.e. user says they cannot login, admin can see that roles are missing. However, this info is also (somewhat implicitly) provided in the frontend now.
Up to you, I can change it to debug.

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
GChuf: The idea is the same for all log level changes - add more info for administrators (it would be nice to know if someone/something is trying to brute force login). I agree this one can be problematic for the logs, though. I can revert this one, and maybe we can find a better way to log (potential) brute force attempts ...

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
GChuf: changed to debug

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
GChuf: reverted to debug

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/auth/HawtioQuarkusAuthenticator.java
GChuf: reverted to debug

hawtio-system/src/main/java/io/hawt/system/Authenticator.java
GChuf: You can see what I came up with in the last commit. doAuthenticate now returns AuthenticateResult instead of Subject.

hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationFilter.java
grgrzybek: I think that adding new enum value makes `hawtio.noCredentials401` obsolete...

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/filters/HawtioQuarkusAuthenticationFilter.java
grgrzybek: `io.hawt.system.AuthenticateResult.Type` didn't have any explanation comments, so it's hard to recall the meaning of these enum values.
You've introduced the `FORBIDDEN` option, but I don't really see a difference with `NOT_AUTHORIZED`... Shouldn't we keep it simple?
* authentication = checking credentials, so if the credentials are wrong, user can provide better ones after getting `401` (show credentials dialog again)
* authorization = checking permissions for authenticated subject/user, so if there's no permission (no proper roles) user can't do anything and `403` is _final_ (don't show credentials dialog again)

?

GChuf: I'm not sure I understand you completely, so bear this in mind while reading my response -

Introducing `FORBIDDEN` makes a lot of sense to me to keep authentication and authorization separate (and/or more clear?), and to align with mozilla's docs referenced above. You mention `401` for unauthenticated, `403` for unauthorized, so that makes me think that we're on the same page, but you also say that you don't see a difference between forbidden and unauthorized. Clearly we can both see the difference between 401 and 403, so isn't this the same?

Returning both codes, while only using `NOT_AUTHORIZED` AuthenticateResult feels wrong to me - although I _just_ realized `AuthenticateResult` implies authentication only, not authorization ... but maybe this one's out of scope :)

I can see your idea of `403` being final. But it was also important to me to give the user and/or the admin a bit more info for failed login attempts.

grgrzybek: I simply treat:
* forbidden == unauthorized (i.e., authorization failed) - that's why I mentioned I don't see a difference ;)
* forbidden != unauthenticated (i.e., authentication failed)

I feel there's something wrong and your intuition is correct. But also remember that the `io.hawt.web.auth.AuthenticationFilter` name is misleading:
* it also performs authorization
* for OIDC and JWT tokens it _only_ does authorization...

hawtio-system/src/main/java/io/hawt/web/ServletHelpers.java
tadayosi: OK, now `NONE` is renamed to `FORBIDDEN`. Let's see if it works ok.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/auth/HawtioQuarkusAuthenticator.java
tadayosi: ok, lets keep this logging.

platforms/quarkus/runtime/src/main/java/io/hawt/quarkus/filters/HawtioQuarkusAuthenticationFilter.java
GChuf: I see what you mean now. I think maybe the **_names_** of the http status codes is the only issue here ... 

This is how I see it:
As mentioned in mozilla docs, unauthorized means "no valid credentials provided"
Therefore:

- forbidden != unauthorized (forbidden: valid credentials were provided but user has no access)
- unauthenticated == unauthorized (valid credentials were not provided)

That said ...
- I agree with your `forbidden == unauthorized` statement more than I do with `forbidden != unauthorized`
- Regardless of how we solve this, I'd like to clearly separate "invalid credentials" vs "valid credentials and invalid access" scenarios.
- re-reading some things maybe the initial login page should return 401 and not 403
- I did not expect this to become complex

grgrzybek: if the initial login page gets 401 and `WWW-Authenticate: Basic realm="xxx"`, then we'd get native browser dialog for entering basic credentials... And we definitely don't want to do that.
We even have special treating of this situation when you "Connect" to external Jolokia agent with basic authentication enabled, where the proxy servlet translates:
```
`WWW-Authenticate: Basic realm="xxx"`
```
to:
```
`WWW-Authenticate: Hawtio original-scheme="Basic" realm="xxx"`
```
so we never get the native credentials dialog.

and your:
> maybe the initial login page should return 401 and not 403

use case is exactly why `noCredentials401` option was created ;)

https://github.com/hawtio/hawtio/blob/d523ee52934b6901c265415c6d6a8e52770d23b1/hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationFilter.java#L149-L157

GChuf: Sorry, busy days. I guess I skipped over that comment regarding the popup. The fact that there's basic auth behind all this also escaped me. So - 403 makes sense in this scenario.
What about the PR itself, what do you think about everything at this point?

hawtio-system/src/main/java/io/hawt/web/auth/AuthenticationFilter.java
tadayosi: For me, this part is the biggest obstacle to approving this PR.

GChuf: I reverted this change now.

PR #4620, opened 2026-07-28
fix(ci): filter audit workflow to report only high and critical severity findings in direct and transient production dependencies

kariuwu:
`--recursive` = audits all direct and transitive dependencies
`--severity high` = outputs high and critical vulnerabilities
`--environment production` = excludes devDependencies (incl. their transitive dependencies)

### DIFF
--- a/.github/workflows/audit-downstream-branches.yml
+++ b/.github/workflows/audit-downstream-branches.yml
@@ -50,4 +50,4 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-${{ matrix.modules }}-
       - name: Audit
-        run: yarn npm audit --all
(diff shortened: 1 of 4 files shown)


kariuwu:
I checked the jobs on my branch first - https://github.com/kariuwu/hawtio/actions/runs/30368630372/job/90306349396

tadayosi (commented):
All look good, except that `--recursive` seems to audit not only the entire hawtio-react workspace (`--all`) but also its dependencies themselves in node_modules; that seems to be a bit excessive.

Is that what you really want to do with this?

tadayosi (approved):
I'm not necessarily opposed to `--recursive`. Let's continue discussing the option elsewhere.

PR #4662, opened 2026-08-04
feat(ci): make an automated slack notfiication that operates when a dowsntream audit fails

kariuwu:
Adding a _notify-team workflow_ that will wait until the _downstream audit workflow_ finishes and sends a notification to our slack channel if there were failures using **Hawtio Audit notification** app

### DIFF
--- a/.github/workflows/audit-downstream-branches.yml
+++ b/.github/workflows/audit-downstream-branches.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        branch: [ 4.x-redhat, 4.4.x-redhat ]
+        branch: [ kariuwu-notify-audit ]
         modules: [ console, console-minimal, examples/sample-plugin ]
     # Navigates into the module directory for all 'run' steps
     defaults:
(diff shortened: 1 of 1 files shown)


kariuwu:
Checked the functionality of the action on my branch first: https://github.com/kariuwu/hawtio/actions/runs/30820813940

.github/workflows/audit-downstream-branches.yml
```diff
@@ -51,3 +51,55 @@ jobs:
...
+                  "text": {
+                    "type": "plain_text",
+                    "text": "Audit Pipeline Failed"
+                  }
+                },
+                {
+                  "type": "section",
+                  "fields": [
+                    {
+                      "type": "mrkdwn",
```
tadayosi, line 78: is this type correct? just checking.

tadayosi (approved):
that's great.
but I'd also recommend for the team to subscribe to the activities on the repo in general via email.

.github/workflows/audit-downstream-branches.yml
kariuwu: It is specific to using slack api: https://docs.slack.dev/tools/slack-github-action/sending-data-slack-incoming-webhook/#usage

kariuwu:
> that's great. but I'd also recommend for the team to subscribe to the activities on the repo in general via email.

Yes, it can be done if the team chooses to watch hawtio repo first and then configure it through the built-in github notifications, but I am guessing it was too inflexible

PR #4671, opened 2026-08-13
chore(deps): Update selenium, selenide, slf4j, junit platform commons.

jsolovjo:
Update Selenium, Selenide and related dependenies to resolve the following issues happening in our CI/CD
```
org.openqa.selenium.devtools.DevToolsException: You are using a no-op implementation of the CDP. 
The most likely reason for this is that Selenium was unable to find an implementation of the CDP protocol that matches your browser. 
Please be sure to include an implementation on the classpath, possibly by adding a new (maven) dependency of `org.seleniumhq.selenium:selenium-devtools-vNN:4.40.0` where `NN` matches the major version of the browser you're using.
```

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -17,13 +17,13 @@
   <properties>
     <http5-version>5.4.3</http5-version>
     <slf4j-version>2.0.11</slf4j-version>
-    <selenium-version>4.40.0</selenium-version>
-    <selenide-version>7.13.0</selenide-version>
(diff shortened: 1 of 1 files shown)


grgrzybek (approved):
Is this the reason of previous SLF4J 2.0.11/2.0.18 failures?

I was on PTO yesterday but I promised to check what's the problem with this update...

grgrzybek (approved):

